PR #1, opened 2021-06-23
Fix up demo

egordon:
First time successfully running the base demo in a while, and first time from this branch, just need to clean up some small things.


amalnanavati (approved):
LGTM

PR #2, opened 2021-07-22
Docs and Sim Updates

amalnanavati:
- Replaced all but one reference to ada_demos with ada_feeding
- Added a Dockerfile for running the demo in sim
- Added README instructions for working with Docker containers
- Clarified a prompt in the sim instructions (it is asking for food item number, not food item name)


egordon (approved):
Thanks for adding the Docker instructions! Hugely beneficial!

PR #3, opened 2021-08-12
Streamlined ADA Actions

amalnanavati:
The existing code loaded *every* parameter that an action or subaction needed at the top level, and passed them through all the actions and subactions that needed them. This makes the code hard to read and understand, for the following reasons:

- Of the parameters that are passed into an action, it is unclear which are actually expected to change across calls to the function, versus which are hardcoded from the beginning.
- Of the parameters that are passed into an action, it is unclear which are actually used in the function body, versus those that are passed onto child functions.
- When developers want to trace where a parameter originally came from, they have to trace it up the whole action call stack.

To address those issues, this PR refactors how parameters are treated within actions’ interfaces and bodies. The changes in this PR follow these guiding principles:

- Every parameter passed into action functions should either: (a) be a parameter that we expect to change across calls to the action; or (b) be a parameter that is expensive to evaluate, and evaluating it earlier in the call stack makes the code more efficient.
- Parameters (e.g., from FeedingDemo) that are inexpensive to access should be loaded at the latest point possible in the call stack.
- Parameters (e.g., from FeedingDemo) that are expensive to access and are used across multiple actions (e.g., the `plate`, which requires some matrix transformations to compute) can be loaded earlier in the call stack.

It is important to acknowledge that although this new system is more readable (for the above reasons), it is less flexible. Earlier, if developers wanted to change parameters across action calls (for example, run an action without velocity limits or without collision constraints), they could do so very easily by passing a different parameter value into the action. Now, if developers want to take a fixed parameter and make it a variable parameter, they will also have to change the function interface. 

Note that this lack of flexibility can be good or bad. On the good side, it requires developers to be very intentional when they are changing a parameter, and requires them to ask: (a) “Should I really be changing this parameter from its default fixed value?” and (b) “If so, should that be done at the level of the function’s interface, or at the level of the config file?” On the bad side, it does require more work for developers who want to change fixed parameters into variable parameters, because they will have to first make that parameter one that is passed into the function.

(Because there are pros and cons to both approaches, I am overall okay with either sticking with the old approach or changing to the approach in this PR. It would be great to have another pair of eyes on this to discuss which approach is better, given our anticipated future steps with the ADA system.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ xauth nlist $DISPLAY | sed -e 's/^..../ffff/' | xauth -f $XAUTH nmerge -
 ```
 8. Run the Docker container using `sudo docker run --rm --network host -v my_catkin_workspace:/workspace -v $XSOCK:$XSOCK:rw -v $XAUTH:$XAUTH:rw -e XAUTHORITY=${XAUTH} -e DISPLAY  -it ada-sim`
 
-You are good to go! Note that any files created within the Docker container that are not within a mounted volume will be deleted once you exit the container (the relevant mounted volume in the above command is your catkin workspace). Further note that the best practice is to have a separate workspace for use within the Docker container, and to only build/source that workspace from within the container (to avoid clashes with ROS versions, library paths, etc.). Finally, note that step 8 above has to be re-run whenever you want to enter your container, and step 7 may have to be re-run.
+You are good to go! Note that any files created within the Docker container that are not within a mounted volume will be deleted once you exit the container (the relevant mounted volume in the above command is your catkin workspace). Further note that the best practice is to have a separate workspace for use within the Docker container, and to only build/source that workspace from within the container (to avoid clashes with ROS versions, library paths, etc.). Finally, note that steps 7-88 above have to be re-run whenever you want to enter your container.
 
 ### PRL Git Packages
 
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ xauth nlist $DISPLAY | sed -e 's/^..../ffff/' | xauth -f $XAUTH nmerge -
 ```
 8. Run the Docker container using `sudo docker run --rm --network host -v my_catkin_workspace:/workspace -v $XSOCK:$XSOCK:rw -v $XAUTH:$XAUTH:rw -e XAUTHORITY=${XAUTH} -e DISPLAY  -it ada-sim`
 
-You are good to go! Note that any files created within the Docker container that are not within a mounted volume will be deleted once you exit the container (the relevant mounted volume in the above command is your catkin workspace). Further note that the best practice is to have a separate workspace for use within the Docker container, and to only build/source that workspace from within the container (to avoid clashes with ROS versions, library paths, etc.). Finally, note that steps 7-88 above have to be re-run whenever you want to enter your container.
+You are good to go! Note that any files created within the Docker container that are not within a mounted volume will be deleted once you exit the container (the relevant mounted volume in the above command is your catkin workspace). Further note that the best practice is to have a separate workspace for use within the Docker container, and to only build/source that workspace from within the container (to avoid clashes with ROS versions, library paths, etc.). Finally, note that steps 7-8 above have to be re-run whenever you want to enter your container.
 
 ### PRL Git Packages
 
--- a/include/feeding/action/DetectAndMoveAboveFood.hpp
+++ b/include/feeding/action/DetectAndMoveAboveFood.hpp
@@ -13,19 +13,10 @@ namespace feeding {
 namespace action {
 
 std::unique_ptr<FoodItem> detectAndMoveAboveFood(
-    const std::shared_ptr<ada::Ada>& ada,
-    const aikido::constraint::dart::CollisionFreePtr& collisionFree,
     const std::shared_ptr<Perception>& perception,
     const std::string& foodName,
-    double heightAboveFood,
-    double horizontalTolerance,
-    double verticalTolerance,
     double rotationTolerance,
-    double tiltTolerance,
-    double planningTimeout,
-    int maxNumTrials,
-    const Eigen::Vector6d& velocityLimits,
-    FeedingDemo* feedingDemo = nullptr,
+    FeedingDemo* feedingDemo,
     double* angleGuess = nullptr,
     int actionOverride = -1);
 }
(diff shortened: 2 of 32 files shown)


README.md
```diff
@@ -51,7 +51,7 @@ xauth nlist $DISPLAY | sed -e 's/^..../ffff/' | xauth -f $XAUTH nmerge -
...
 
-You are good to go! Note that any files created within the Docker container that are not within a mounted volume will be deleted once you exit the container (the relevant mounted volume in the above command is your catkin workspace). Further note that the best practice is to have a separate workspace for use within the Docker container, and to only build/source that workspace from within the container (to avoid clashes with ROS versions, library paths, etc.). Finally, note that step 8 above has to be re-run whenever you want to enter your container, and step 7 may have to be re-run.
+You are good to go! Note that any files created within the Docker container that are not within a mounted volume will be deleted once you exit the container (the relevant mounted volume in the above command is your catkin workspace). Further note that the best practice is to have a separate workspace for use within the Docker container, and to only build/source that workspace from within the container (to avoid clashes with ROS versions, library paths, etc.). Finally, note that steps 7-88 above have to be re-run whenever you want to enter your container.
```
egordon, line 54: Typo: 7-8, not 7-88

egordon (commented):
I'll go over parameters more closely next week.

As for you concern about flexibility: I think a good approach we should take in the future is to bake the flexibility into the FeedingDemo object (and maybe refactor it to be a "FeedingParams" object instead) and the YAML configuration file.

As discussed previously, every action could identify itself with a string (e.g. "moveAbove"). And this can be a key in the YAML file. The default key is "default". Then, we can add an interface to FeedingDemo to retrieve a parameter with the given key. This can reintroduce flexibility back into the system without sacrificing readability.

I don't think we should add this to this particular PR, but I added it as an item in the Infra Tasks list to discuss later.

README.md
amalnanavati: Done

include/feeding/action/DetectAndMoveAboveFood.hpp
```diff
@@ -13,22 +13,13 @@ namespace feeding {
...
-    double verticalTolerance,
     double rotationTolerance,
-    double tiltTolerance,
-    double planningTimeout,
-    int maxNumTrials,
-    const Eigen::Vector6d& velocityLimits,
-    FeedingDemo* feedingDemo = nullptr,
+    FeedingDemo* feedingDemo,
     double* angleGuess = nullptr,
     int actionOverride = -1);
```
egordon, line 21: Commenting on the actionOverride here. It was in place for study-purposes where we can force the robot to choose an action that was not recommended by SP/SPANet.

In the future, I would be in favor of separating the "Detect" from the "MoveAbove". As the fact that they have been merged by default has made it difficult to run experiments. Then, the override could be handled at a higher level.

include/feeding/action/FeedFoodToPerson.hpp
```diff
@@ -12,29 +12,10 @@ namespace feeding {
...
 
 void feedFoodToPerson(
-    const std::shared_ptr<ada::Ada>& ada,
-    const std::shared_ptr<Workspace>& workspace,
-    const aikido::constraint::dart::CollisionFreePtr& collisionFree,
-    const aikido::constraint::dart::CollisionFreePtr&
-        collisionFreeWithWallFurtherBack,
     const std::shared_ptr<Perception>& perception,
     const ros::NodeHandle* nodeHandle,
     const Eigen::Isometry3d& plate,
```
egordon, line 17: Note for later that I'm not sure why we would need the plate transform here, except that is is necessary for the final MoveAbovePlate action (if that's in here). If it is, we should separate it out.

See below comment as well, but I think we should have separate "FeedFoodToPerson", "Tilt Food", and "Return To Plate" actions, which should make this easier to understand.

include/feeding/action/FeedFoodToPerson.hpp
```diff
@@ -12,29 +12,10 @@ namespace feeding {
...
-    double rotationToleranceAbovePlate,
-    double distanceToPerson,
-    double horizontalToleranceForPerson,
-    double verticalToleranceForPerson,
-    double planningTimeout,
-    int maxNumTrials,
-    double endEffectorOffsetPositionTolerenace,
-    double endEffectorOffsetAngularTolerance,
-    const Eigen::Vector6d& velocityLimits,
     const Eigen::Vector3d* tiltOffset,
```
egordon, line 19: See above, I think a separate Tilt action can make more sense here.

include/feeding/action/MoveTowardsPerson.hpp
```diff
@@ -3,21 +3,19 @@
...
 #include "feeding/perception/Perception.hpp"
 
 namespace feeding {
 namespace action {
 
 bool moveTowardsPerson(
-    const std::shared_ptr<ada::Ada>& ada,
     const aikido::constraint::dart::CollisionFreePtr& collisionFree,
     const std::shared_ptr<Perception>& perception,
     const ros::NodeHandle* nodeHandle,
```
egordon, line 16: Does the FeedingDemo not have the NodeHandle?

egordon (approved):
Left some comments on individual parameters, nothing that will block merging.

That said, I think we should create a second PR soon to document all of the remaining arguments, especially the ones that aren't obvious (e.g. `angleGuess` and `plate` in FeedFoodToPerson)

Added as an issue: https://github.com/personalrobotics/ada_feeding/issues/4

include/feeding/action/DetectAndMoveAboveFood.hpp
amalnanavati: Agreed on separating Detect and MoveAbove. I'll keep action override as-is for this PR, but when we separate those two we should decide where to best place the override.

include/feeding/action/FeedFoodToPerson.hpp
amalnanavati: Yup, this behemoth of an action does the whole bite transfer and then returns to the plate. Agreed that "Return to Plate" should be separated out. 

I'm not sure whether "Tilt Food" should be separated from "FeedFoodToPerson" because imo it is all part of the bite transfer and the reason to tilt the food is to facilitate feeding it to the person. Further, I could envision bite transfer trajectories where the timing of tilting matters, and therefore you can't tilt the food independently from the bite transfer.

Regardless, we can/should discuss this later for when we actually split off the actions.

include/feeding/action/MoveTowardsPerson.hpp
amalnanavati: Good catch. Removed nodeHandle from all action interfaces

PR #8, opened 2021-11-08
Integration with new libada + making ada_feeding hardware agnostic to gen2 & gen3

RKJenamani:
- Integrated with new libada (egordon/simplify)
    - Removed dependence on libada/resources/tsr_transforms.yaml (transforms are hardcoded now)
    - Removed dependence on ada_launch
    - Changed moveArmTo calls to planTo + executeTrajectory calls
    - Removed functions from utils that are not useful anymore
- Made ada_feeding hardware agnostic
    - Added separate configuration files for gen2 & gen3
    - Current state works on gen2 real (ToDo: check on gen3)


CMakeLists.txt
```diff
@@ -34,8 +33,10 @@ find_package(aikido 0.3.0 REQUIRED
     planner_vectorfield
     planner_kunzretimer
     robot
-    rviz
+    robot_ros
     statespace
+    io
```
amalnanavati, line 38: Hmm, was there a reason for moving `io`? I believe the norm is to alphabetize it (in which case, rviz comes before statespace and io after distance)

include/feeding/util.hpp
```diff
@@ -37,8 +37,9 @@ static const std::vector<std::string> FOOD_NAMES = {"apple",
                                                     "spinach",
                                                     "kale"};
 
+// TODO(egordon): Reference from YAML file
```
amalnanavati, line 40: Hmm, isn't FOOD_NAMES in the yaml, and can't we add BEST_ACTIONS to the yaml? That might be what this TODO was referring to.

scripts/bite_location_detector.cpp
```diff
@@ -152,8 +152,13 @@ void bite_location_detector(
       aboveFoodTSR.mT0_w = foodTransform;
       aboveFoodTSR.mTw_e.translation() = Eigen::Vector3d{0, 0, -0.03};
       aboveFoodTSR.mBw = createBwMatrixForTSR(0.001, 0.001, 0, 0);
-      Eigen::Isometry3d eeTransform
-          = *feedingDemo.getAda().getHand()->getEndEffectorTransform("food");
+      // TODO: remove hardcoded transform for food
```
amalnanavati, line 155: Flagging this TODO. Do we want to address it in this PR?

scripts/main.cpp
```diff
@@ -138,7 +138,8 @@ int main(int argc, char** argv)
   if (ftThresholdHelper)
     ftThresholdHelper->init();
 
-  feedingDemo->getAda()->closeHand();
+  // TODO: uncomment this once we can send actions to hand
```
amalnanavati, line 141: I haven't had any issues even with this line commented out. Should we just leave it out (or if not, what does it add)?

src/FeedingDemo.cpp
```diff
@@ -63,43 +62,43 @@ FeedingDemo::FeedingDemo(
...
-          mWorkspace->getWheelchair().get());
-
-  mCollisionFreeConstraintWithWallFurtherBack
-      = std::make_shared<aikido::constraint::dart::CollisionFree>(
-          mArmSpace,
-          mAda->getArm()->getMetaSkeleton(),
-          relaxedCollisionDetector);
-  mCollisionFreeConstraintWithWallFurtherBack->addPairwiseCheck(
-      relaxedArmCollisionGroup, relaxedEnvCollisionGroup);
+  // dart::collision::CollisionDetectorPtr collisionDetector
```
amalnanavati, line 65: Should we just remove this long block of comments, now that we have verified that the new collision checking works with the feeding demo?

src/FeedingDemo.cpp
```diff
@@ -286,9 +306,13 @@ std::shared_ptr<FTThresholdHelper> FeedingDemo::getFTThresholdHelper()
 //==============================================================================
 Eigen::Isometry3d FeedingDemo::getPlateEndEffectorTransform() const
 {
-  Eigen::Isometry3d eeTransform
-      = mAda->getHand()->getEndEffectorTransform("plate").get();
-  eeTransform.linear() = eeTransform.linear()
+  // TODO: remove hardcoded transform for plate
```
amalnanavati, line 309: Flagging this as well

src/action/FeedFoodToPerson.cpp
```diff
@@ -238,13 +247,34 @@ void feedFoodToPerson(
...
     talk("Tilting, hold tight.", true);
 
-    ada->moveArmToTSR(
-        personTSR,
-        nullptr, // collisionFreeWithWallFurtherBack,
-        planningTimeout,
-        maxNumTrials,
-        getConfigurationRanker(ada),
-        slowerVelocity);
+    auto trajectory = ada->getArm()->planToConfiguration(ada->getArm()->getNamedConfiguration("home_config"),ada->getArm()->getSelfCollisionConstraint());
```
amalnanavati, line 250: Huh. I didn't see an issue when running the code, but why does it move to the home config first before moving to the `personTSR`? Why not do it all in one go as the original code does?

src/action/MoveAboveFood.cpp
```diff
@@ -33,8 +33,13 @@ bool moveAboveFood(
   // double tiltTolerance = feedingDemo->mFoodTSRParameters.at("tiltTolerance");
 
   Eigen::Isometry3d target;
-  Eigen::Isometry3d eeTransform
-      = *ada->getHand()->getEndEffectorTransform("food");
+  // TODO: remove hardcoded transform for food
```
amalnanavati, line 36: Flagging this TODO as well

src/action/MoveAboveFood.cpp
```diff
@@ -67,6 +72,8 @@ bool moveAboveFood(
   {
     eeTransform.linear() = eeTransform.linear() * rotation;
     eeTransform.translation()[2] = heightAboveFood;
+    eeTransform.translation()[0] += 0.01;
```
amalnanavati, line 75: FYI, these fudge factors can be removed once the camera is properly calibrated and the fork properly matches the URDF. With gen2 real, we were able to remove these fudge factors by: (a) properly [calibrating the camera](https://github.com/personalrobotics/libada/pull/67); and (b) bending the fork tines to undo the deformations it had faced over the years.

In terms of the code itself, I have a marginal preference for removing the fudge factors in the `main` branch and adding then in on any local version of the code that needs it. If others agree with this position, I can push a commit with the fudge factors commented out.

src/action/MoveAboveForque.cpp
```diff
@@ -35,11 +35,34 @@ void moveAboveForque(
   aboveForqueTSR.mT0_w = forquePose;
 
   aboveForqueTSR.mBw = createBwMatrixForTSR(0.0001, 0.0001, 0.0001, 0, 0, 0);
+
+  // TODO: Remove hardcoded transform for plate
```
amalnanavati, line 39: Flagging this TODO as well

src/action/MoveAbovePlate.cpp
```diff
@@ -17,29 +17,19 @@ bool moveAbovePlate(
...
-  homeConfig << -2.11666, 3.34967, 2.04129, -2.30031, -2.34026, 2.9545;
-  bool success = ada->moveArmToConfiguration(
-      homeConfig, collisionFree, 2.0, velocityLimits);
-  if (!success)
+  auto trajectory = ada->getArm()->planToConfiguration(ada->getArm()->getNamedConfiguration("home_config"),ada->getArm()->getWorldCollisionConstraint());
+  bool success = true;
+  auto future = ada->getArm()->executeTrajectory(trajectory); // check velocity limits are set in FeedingDemo
+  try
   {
-    auto retval = moveAbove(
```
amalnanavati, line 29: Interesting that `MoveAbovePlate` no longer calls `moveAbove`, but `MoveAboveFood` does. Why the discrepancy? (Understanding the utility or lack thereof of `moveAbove` versus the other actions can help us as we think about refactoring the actions/subactions)

src/action/MoveDirectlyToPerson.cpp
```diff
@@ -49,8 +49,14 @@ bool moveDirectlyToPerson(
         0,
         M_PI / 4,
         M_PI / 4);
-    Eigen::Isometry3d eeTransform
-        = ada->getHand()->getEndEffectorTransform("person").get();
+
+    // TODO: remove hardcoded transform for person
```
amalnanavati, line 53: Flagging this TODO

src/action/MoveDirectlyToPerson.cpp
```diff
@@ -67,8 +73,15 @@ bool moveDirectlyToPerson(
         0,
         0,
         0);
+    // TODO: Remove hardcoded transform for person
```
amalnanavati, line 76: Ditto

src/action/MoveInFrontOfPerson.cpp
```diff
@@ -63,16 +62,37 @@ bool moveInFrontOfPerson(
       0,
       0,
       0);
+
+  // TODO: Remove hardcoded transform for person
```
amalnanavati, line 66: Flagging this TODO

src/action/Skewer.cpp
```diff
@@ -212,7 +212,9 @@ bool skewer(
           forkXAxis[2] = 0.0;
           forkXAxis.normalize();
           endEffectorDirection *= heightAboveFood;
-          endEffectorDirection += ((-0.02 * forkYAxis) + (-0.005 * forkXAxis));
+          // TODO: add the coefficients below as parameters to demo config for compatibility with Gen2/3
+          // Note: Should be (-0.02, -0.005) for Gen2
+          endEffectorDirection += ((0.0 * forkYAxis) + (0.0 * forkXAxis));
```
amalnanavati, line 217: Ditto the comment above about fudge factors -- this is no longer necessary for gen2 due to the [camera calibration code](https://github.com/personalrobotics/libada/pull/67).

src/perception/Perception.cpp
```diff
@@ -195,16 +195,9 @@ Eigen::Isometry3d Perception::perceiveFace()
       // fixed distance:
       double fixedFaceY
           = getRosParam<double>("/feedingDemo/fixedFaceY", *mNodeHandle);
-      if (fixedFaceY > 0)
-      {
-        faceTransform.translation().y() = fixedFaceY;
-        // Wheelchair
-        faceTransform.translation().z() -= 0.02;
+      faceTransform.translation().y() = fixedFaceY;
+      faceTransform.translation().z() -= 0.01;
```
amalnanavati, line 199: I'm guessing this is the out-of-mouth direction? Why do we adjust it by 1cm, given that we already have `distanceToPerson` as a param in the config file?

amalnanavati (commented):
Overall LGTM. There are a few minor comments to be addressed/discussed, but this is close to merging.

src/perception/Perception.cpp
egordon: This is not the out-of-mouth direction (that is addressed by `fixedFaceY`). This is the vertical direction (against gravity).

I believe the issue is that when we plan towards the stomion, a user would somehow need to move their head perfectly forward to get the food, but that's not how the neck (usually) works.

So we plan slightly *below* the stomion, so when the user bends their neck, they can more comfortably get the food.

Formalizing this as a parameter can be a separate issue.

amalnanavati: Ooh that makes a lot of sense! I think it could be useful to add a comment on this (for whoever next makes a push on this branch)

src/action/MoveAbovePlate.cpp
amalnanavati: Because `MoveAbovePlate` moves to a fixed configuration, and doesn't actually move above a fixed position.

include/feeding/util.hpp
amalnanavati: I believe Ethan addressed this

amalnanavati:
Regarding all the TODOs re. the hardcoded rotation angles, Ethan and I discussed that going through and changing those should be in another PR, not this one. Hence, I resolved them.

It seems like there are two lingering discussions (although they probably don't require changes) and then this should be ready to merge in.

src/ranker/ShortestDistanceRanker.cpp
```diff
@@ -11,52 +11,34 @@ namespace feeding {
...
-        break;
-      case 2:
-        tiltStyle = TiltStyle::ANGLED;
-        break;
-      default:
-        tiltStyle = TiltStyle::NONE;
-    }
-    rotation = (actionNum % 2 == 0) ? 0.0 : M_PI / 2.0;
+  // Get Action From DetectedObject
+  std::string itemAction = item.getInfoByKey<std::string>("action");
```
amalnanavati, line 27: @egordon If I'm understanding this correctly, the code now uses the action and rotation that is outputted by food detection. Is there no longer any way to override it, particularly the tiltStyle, using the YAML? (Because if you recall, even the code earlier had hardcoded best actions -- it seems like the new code has no way to override the action or tilt style outputted by perception).

amalnanavati (approved):

src/action/Skewer.cpp
```diff
@@ -13,114 +13,98 @@
...
   std::chrono::milliseconds waitTimeForFood = feedingDemo->mWaitTimeForFood;
   // const Eigen::Vector6d& velocityLimits = feedingDemo.mVelocityLimits;
-  const std::shared_ptr<FTThresholdHelper>& ftThresholdHelper = feedingDemo->getFTThresholdHelper();
-  std::vector<std::string> rotationFreeFoodNames = feedingDemo->mRotationFreeFoodNames;
+  const std::shared_ptr<FTThresholdHelper> &ftThresholdHelper =
+      feedingDemo->getFTThresholdHelper();
+  std::vector<std::string> rotationFreeFoodNames =
+      feedingDemo->mRotationFreeFoodNames;
+
+  // TODO (egordon): add option to disable override in feeding_demo.yaml
```
egordon, line 45: @amalnanavati This is where the YAML overrides the perception. This variable `actionOverride` was initially set to -1, and only used in the user study if the user explicitly selected an action.

Now, this variable is set based on the YAML (if it exists), and overrides perception.

src/ranker/ShortestDistanceRanker.cpp
egordon: See below comment, the override now happens on a higher level rather than in the Ranker.

This also means that now all Rankers (ShortestDistance and SuccessRate) function identically, pulling the action from perception to be overridden (optionally) later.

PR #9, opened 2021-12-14
Improved the Audio Interaction for the Demo

amalnanavati:
- Added a useSound command line flag that controls whether to use speech output or not.
- Made most speech output run in the background (i.e., non-blocking)
- Added a useAlexa parameter that dictates whether to wait to receive a food name from Alexa, or just ask for command line input
- Slightly changed the speech outputs over the course of the demo

The merge order should be:

1. `emprise/gen3-integration` --> `main`
2. `amaln/batch_size_params` --> `main` (this should be paired with the corresponding AIKIDO PR)
3. `amaln/audio_improvements` --> `main`

(The second merge is not a strict dependency, but will make merging easier)

### DIFF
--- a/config/gen2_feeding_demo.yaml
+++ b/config/gen2_feeding_demo.yaml
@@ -128,7 +128,7 @@ humanStudy:
   autoTiming: true
   autoTransfer: true
   createError: false
-  useAlexa: true
+  useAlexa: false
   foodTopic: /alexa_msgs
   actionTopic: /alexa_msgs
 
--- a/config/gen2_feeding_demo.yaml
+++ b/config/gen2_feeding_demo.yaml
@@ -121,7 +121,7 @@ foodItems:
   #pickUpAngleModes: [0, 2, 2, 0, 0, 0, 0]
   names: ["apple", "banana", "bell_pepper", "broccoli", "cantaloupe", "carrot", "cauliflower", "celery", "cherry_tomato", "grape", "honeydew", "kiwi", "strawberry", "lettuce", "spinach", "kale"]
   forces: [10, 10, 10, 10, 15, 22, 7, 22, 7, 10, 7, 7, 15, 7, 7, 7]
-  pickUpAngleModes: [1, 5, 1, 2, 1, 1, 0, 1, 1, 3, 3, 3, 0, 1, 2, 0]
+  pickUpAngleModes: [1, 5, 1, 2, 1, 1, 0, 1, 1, 1, 3, 3, 0, 1, 2, 0]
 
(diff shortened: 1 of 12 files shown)


amalnanavati:
The second commit are some additional changes made for the demo. I'm ambivalent about whether they are included in this PR or not -- I can spin off a separate PR if reviewer(s) prefer.

config/gen2_feeding_demo.yaml
```diff
@@ -121,14 +121,14 @@ foodItems:
...
-  pickUpAngleModes: [1, 5, 1, 2, 1, 1, 0, 1, 1, 3, 3, 3, 0, 1, 2, 0]
+  pickUpAngleModes: [1, 5, 1, 2, 1, 1, 0, 1, 1, 1, 3, 3, 0, 1, 2, 0]
 
 humanStudy:
   autoAcquisition: true
   autoTiming: true
   autoTransfer: true
   createError: false
-  useAlexa: true
+  useAlexa: false
```
egordon, line 131: Please add this to gen3_feeding_demo as well, since you don't have a default arg set in FeedingDemo.cpp.

src/action/FeedFoodToPerson.cpp
```diff
@@ -261,12 +269,14 @@ void feedFoodToPerson(const std::shared_ptr<Perception> &perception,
   if (moveIFOSuccess) {
     // ===== EATING =====
     ROS_WARN("Human is eating");
-    talk("Ready to eat!");
+    if (useSound) 
+      talk("Ready to eat!");
```
egordon, line 273: Does ada_feeding use clang_format?
Because I'm pretty sure unbracketed ifs aren't allowed.
This can be pushed as-is right now if there is no clang-format call, and we should add one soon.
If there is a clang-format call, let's do that first.

egordon (changes requested):
Making sure that things still work on Gen3.
And a formatting nit (maybe begetting another issue). Otherwise, looks okay.

config/gen2_feeding_demo.yaml
amalnanavati: Done

src/action/FeedFoodToPerson.cpp
amalnanavati: We don't have clang_format yet. I added an issue for it https://github.com/personalrobotics/ada_feeding/issues/10

amalnanavati:
Ready for re-review

egordon (approved):
LGTM, thanks!

PR #17, opened 2023-02-09
Implementing plate locator algorithm

Raidakarim:


amalnanavati:
I'm guessing there will be no more work on this since it is ROS1. Closing.

PR #19, opened 2023-05-03
Added Custom ROS2 Actions, Services, Messages

amalnanavati:
This PR creates the interfaces for the actions, services, and messages we envision using for the ROS2 robot feeding system.

For an earlier version of our brainstorming on this, check out [this Google Sheet](https://docs.google.com/spreadsheets/d/1px5VTPjxGtOvgd8u3r53WzR5Cc6I-QT2qKyyeGWF-p8/edit#gid=574571123).


amalnanavati:
@egordon @BernieZhu @taylorkf I won't merge this PR until I have (at a minimum) developed dummy nodes for all of these in the web app. But I wanted to tag you all to get the conversation started early about whether these interfaces align with our expectations from the robot, perception, and web app side. It'll be much easier to make interface changes now than downstream :)

amalnanavati:
Heads up, I removed the images from calls to the perception actions because the web app will never be subscribing to a raw image topic. Rather, it will pass a raw image topic into [`web_video_server`](https://github.com/RobotWebTools/web_video_server/tree/ros2) and then render the resulting video stream. As a result of that, the web app will not have the image messages to pass into the perception actions.

Therefore, when perception actions are called, they will have to get the latest image from the camera topic. We could do this in three ways:
1. The node is always subscribed to the camera topic and stores the latest camera image. When the action is called, it uses that. 
    - **Pros**: The latest image is ready when the action is called.
    - **Cons**: The node is always subscribed to the camera topic and storing the latest image, even if we won't be calling the perception action anytime soon.
2. When the action is called, the node subscribes to the image topic and doesn't proceed with the action until it gets one image.
    - **Pros**: The node is not subscribed to the image topic and storing images unless it is actively needed by the web app.
    - **Cons**: The action will take longer.
3. We have a service that toggles the perception action on/off. When it is on, it subscribes to the image topic (scenario 1). When it is off, it doesn't. If the action is called when it is off, it immediately aborts the action.
    - **Pros**: The perception is not subscribed to the image topic when not needed, but will be toggled on before the action so it will have a ready image by the time the action is called.
    - **Cons**: Additional services :)

(NOTE that in all these cases, because the subscriber callback and the action callback will be called in separate threads, we need to protect the latest image variable in a mutex.)

What are your thoughts @BernieZhu @egordon ?

amalnanavati:
I decided to make FaceDetection a subscriber/publisher model, which is toggled on and off by a service. The reason for this is that we use face detection in two places:
1. To wait until the user has opened their mouth. In this case, a `WaitUntilOpenMouth` action can suffice, but what if we want to render annotated images of the results of face detection while waiting? In that case we will need to have it publish anyway.
2. To show the results of the face detection as the user is customizing the staging location. In this case, there is no concrete "end" to face detection, so an action doesn't make sense.

Therefore, I decided to make face detection subscribe to the image stream, and publish both a `FaceDetection` message and an annotated image. It can be toggled on and off by a service.

amalnanavati:
Since all the dummy node implementations in the web app have been approved (feeding_web_interface [#34](https://github.com/personalrobotics/feeding_web_interface/pull/34), [#36](https://github.com/personalrobotics/feeding_web_interface/pull/36), [#37](https://github.com/personalrobotics/feeding_web_interface/pull/37), [#38](https://github.com/personalrobotics/feeding_web_interface/pull/38), [#43](https://github.com/personalrobotics/feeding_web_interface/pull/43)), I am taking that as implicit approval that these interfaces are good and make sense. (Plus, we already discussed and loosely agreed upon these interfaces in ADA steering meetings.) Hence, I'll go ahead and merge.

PR #20, opened 2023-05-24
Generic Wrapper for Executing py_trees within a ROS2 Action Server

amalnanavati:
# Why is this PR necessary?

The `ada_feeding` code should work in (at least) two different modes:
1. **App-Centric**: Each behavior tree is wrapped in a ROS2 action server, which is invoked by the app.
2. **Robot-Centric**: The feeding system runs as one unified behavior tree that sometimes waits for particular inputs (which may or may not be from the app).

However, to promote modularity and reusability and to prevent bugs, we should ideally re-use the core functionalities across these different nodes. This PR enables that.

# What does this PR change/add?

## `create_action_servers.py`

This PR defines a `create_action_servers.py` ROS2 node that runs behavior trees as ROS2 servers. Specifically, this node takes in ROS Params that specify:
- `server_name`: The name of the action server.
- `action_type`: The ROS2 action type that this action expects.
- `tree_class`: The class containing the behavior tree for this action server to execute. **This class must subclass `ada_feeding.ActionServerBT`**.
- `tree_kwargs`: An arbitrary dictionary of arguments to be used to initialize the instance of `tree_class`.
- `tick_rate`: How often to tick the tree when the action server is executing a goal.
- 
`create_action_servers.py` then creates the action servers specified in the params. When an action server is called, it loads the corresponding tree and starts ticking it at `tick_rate`, publishing feedback and the result as specified in the `tree_class`.

## `ActionServerBT`

This PR defines an abstract class, `ada_feeding.ActionServerBT`, that connects ROS2 actions and py_trees. Specifically, this class has to do the following:
- `create_tree`: Create the tree. Called **once**, sometime before the first call to this action server begins executing.
- `send_goal`: Take a goal from the action server and send it to the tree. Called **once per goal** that is sent to this action server.
- `get_feedback`: Return the feedback type that the ROS2 action is expecting, by examining the tree's current state. Called **once per iteration** of the action server's callback loop.
- `get_result`: Return the result type that the ROS2 action is expecting, by examining the tree's current state. Called **once per goal** after the tree's status has switched away from `RUNNING`.

By design, subclasses of `ActionServerBT` are the only classes that should need to explicitly import specific ROS2 action types. The tree should store the information needed for those action messages in its blackboard, but it doesn't need to have access to the specific ROS2 action class. And `create_action_servers.py` should import whatever action types are specified in the config file.

## Dummy Behaviors/Trees

Finally, this PR implements a series of dummy behaviors and tree files that use those behaviors, in order to test the code and ensure that it all works as expected with the web app. These dummy implementations include:
- Behaviors:
    - `move_to_dummy.py`: Meant to simulate a behavior that might call MoveIt2's MoveGroup action. This dummy behavior sleeps for a set amount of time to simulate "planning" and another set amount of time to simulate "motion" before returning.
- Trees:
    - `move_above_plate_dummy.py`: This single-behavior tree uses `move_to_dummy.py` to mimic the move above plate action. It serves as a working example of subclassing `ActionServerBT`.

# Design decisions

- **One goal at a time**: The entire `create_action_servers.py` node will only accept one goal at a time. This is because we intend this node to expose all the actions related to robot motion, but not actions related to other aspects of the system such as perception. Since we will primarily use this node for robot motion, it makes sense that only one action can be executing on the robot at a time.
- **tick_tock**: I didn't use py_tree's built-in `tick_tock` feature because that is intended for a tree that ticks indefinetely, whereas we want the tree to stop ticking when the action succeeds, fails, or is preempted. Although it is possible to interrupt a tree's `tick_tock`, it is unclear to me whether that can/should be done from the pre/post-tick handlers (which is what we'd need, e.g., pre-tick check if a preempt has been requested and interrupt the `tick_tock`). Hence, I implement a custom tick loop in the `execute_callback` function of the action servers.
- **Pre/Post-tick handlers**: Because I don't use `tick_tock`, the use of pre- and post-tick handlers is no longer necessary, and I instead call those functions directly before running `tree.tick()`.
- **Preemption handling**: ROS2 action server preemptions are handled by calling `tree.stop(INVALID)`, which is the same pathway that is used if a higher-order behavior in a tree cancels a lower-order behavior. Internally, this ends up calling `terminate` on every behavior in the tree, so `terminate` must properly implement stopping the behavior (e.g., preempting any ROS actions the behavior called). Note that the ROS action execution thread blocks until the tree has processed the termination.
    - Note that from the perspective of an action client, calling `client.cancel_goal(...)` is not enough, because that returns when the cancellation request is accepted. To identify when the cancellation request has been processed, you must use `goal_handle.get_result().status`.

# Testing procedure
1. Build the workspace: `colcon build`
2. Source the workspace: `source install/setup.bash`
3. Run the action server(s): `ros2 launch ada_feeding ada_feeding_launch.xml`
4. Test the code with each action (the below code is for `MoveAbovePlate` but the test was done for each action):
    - [x] Call `MoveAbovePlate`, ensure it runs to completion and returns the expected feedback: `ros2 action send_goal /MoveAbovePlate ada_feeding_msgs/action/MoveTo "{}" --feedback`
        - [x] Run `ps-ax` before sending a goal and ensure there is only one `MoveAbovePlate` process running. Then, call it while a goal is executing, and ensure there are two. Then, call it after the goal has ended (either in success or canceled) and ensure there is again only one process. (This is particular to the dummy node implementation)
    - [x] Once one call to `MoveAbovePlate` has succesfully ended, call it again (above command) and ensure that works.
    - [x] Call `MoveAbovePlate` (above command), and terminate the call, and ensure the cancel is handled as expected.
    - [x] Once one call to `MoveAbovePlate` has been canceled, call it again (above command) and ensure that works.
    - [x] Call `MoveAbovePlate` (above command) and call it again while that is executing, and ensure that the new goal is rejected and the first call behaves as expected.
    - [x] Terminate and then call `MoveAbovePlate` in quick succession and ensure it behaves as expected.
5. Test the code with two actions (`MoveAbovePlate` and `AcquireFood`):
    - [ ] Call `MoveAbovePlate` and, while it is running call `Acquire Food` (`ros2 action send_goal /AcquireFood ada_feeding_msgs/action/AcquireFood "{header: {stamp: {sec: 0, nanosec: 0}, frame_id: ''}, detected_food: {roi: {x_offset: 0, y_offset: 0, height: 0, width: 0, do_rectify: false}, mask: {header: {stamp: {sec: 0, nanosec: 0}, frame_id: ''}, format: '', data: []}, item_id: '', confidence: 0.0}}" --feedback`) and ensure the call to `AcquireFood` gets rejected.
    - [ ] The same test as above but swap the two actions.
    - [ ] After one action has succeeded, call another action, and ensure it works.
    - [ ] Call one action, then cancel it and immediately call another action, ensure it behaves as exected.
6. Test the entire system together:
    - Load the dummy perception nodes: `TBD`
    - Run the web app.
    - [ ] Ensure that the web app works as expected and is calling the movement actions as expected.

# Before Merging
- [x] `Squash & Merge`


egordon:
Great write-up, thank you so much @amalnanavati !

Not approving yet since I haven't done a full review, some comments:
* Maybe consider read-only ros parameters for create_action_servers.py so you can place all params. For kwargs, params now allow [dynamic typing](https://github.com/ros2/rclcpp/blob/rolling/rclcpp/doc/notes_on_statically_typed_parameters.md) and [undeclared parameters](https://docs.ros2.org/foxy/api/rclpy/api/node.html).
* For preemption handling, I believe it is possible to implement this a bit more elegantly with the blocking approach without actually blocking the ROS action server. See [this example](https://github.com/ros2/examples/blob/master/rclpy/actions/minimal_action_server/examples_rclpy_minimal_action_server/server_single_goal.py). Rather than calling the tree's preempt function, just immediately return ACCEPT, this does *not* mean that the goal is cancelled, it just alerts the client that the cancelling is underway. The actual cancellation can happen in the execute loop by calling `goal_handle.canceled()` when the (potentially blocking) preempt function is called. That potentially obviates the need for a `was_preempted()` boolean function.

Everything else looks conceptually fine, thank you again!

egordon:
@amalnanavati My understanding is that `execute_callback` is already run asynchronously from any polling the underlying action server does, so you can take as long as you want there without worrying about holding up other server functions.

After all, your tree execution is within a `while rclpy.ok():` block, so your `execute_callback` is blocking anyway until the tree is done. It is against best BT practices for a *tick* to block, but that's not what's happening here. I would go with option (a).

Additionally, you can consider calling the tree's `shutdown` function (which should by default loop through all nodes and call their respective `shutdown` functions, similarly to `setup`), which could handle tree cleanup without needing to define the `preempt` function, further simplifying the interface.

amalnanavati:
The reason I didn’t go with the shutdown option is so we can cache trees across different calls to the same action server, without re-calling `setup`.  But I don’t mind switching back to that. I’ll look more into what the shutdown does exactly and what is needed to start up again after a shutdown

egordon:
I think caching trees only makes sense if there are resources a given tree has that don't need to be released when a different tree is called and `setup()` would take a long time.

Even then, it may be better to store that cached info in the ActionServerBT object (passed into each new tree created with `create_tree`) instead, an then `setup` and `shutdown` can operate on per-instance operations.

amalnanavati:
@egordon Commit [a0ba5e0](https://github.com/personalrobotics/ada_feeding/pull/20/commits/a0ba5e0e31046cd4895fa1d209a28cf0813ff831)  should address your concern with preemption. In the new formulation, a preemption request from ROS action servers is handled by calling the `terminate` method on every behavior in the tree. This is nice because now the same function that handles cancellations from behaviors higher-up in the tree will also handle preemptions from the ROS action server, which promotes reuse. In this formulation, `preempt_goal(...)` is implemented in `ActionServerBT` itself, so neither the tree nor its behaviors need to explicitly handle ROS action server preemptions. As long as behaviors correctly implement `terminate`, everything should work.

In addition, commit [ebe16c2](https://github.com/personalrobotics/ada_feeding/pull/20/commits/ebe16c2bc94d02e3244823589967bd7c45942eef) addresses your feedback on `setup`/`shutdown`, by carrying it out at the beginning and end of each call to the action server.

Commit [098fef1](https://github.com/personalrobotics/ada_feeding/pull/20/commits/098fef144d6482e2cb50184a396614a295d886cd) implements your feedback on caching. It is no longer the default behavior in `create_action_servers.py`, but can be optionally implemented in each subclass of `ActionServerBT`.

And finally, commit [39ebcd7](https://github.com/personalrobotics/ada_feeding/pull/20/commits/39ebcd77f7299ad9a9a9f3d0460a7b792b5c9634) refactors the config file to use ROS params and makes them all read only and properly typed.

I've updated the PR write-up accordingly. Thanks for the helpful feedback!

amalnanavati:
@taylorkf Can you re-run your test with the new changes? If it still doesn't work, can you document additional details (e.g., how soon after canceling do you re-run) so I can reproduce it? I haven't been able to reproduce the issue thus far.

taylorkf (approved):
I'm happy with the changes that @amalnanavati and @egordon agreed to, the code looks good, and the tests run perfectly on my machine. Nice going!

amalnanavati:
Great, I'll merge this PR in, and then make the additional dummy movement trees (which should be very similar to `MoveAbovePlate`) a separate PR.

PR #25, opened 2023-06-02
Removed MoveToMouth action

amalnanavati:
Now that we have removed the staging location, we no longer need a separate `MoveToMouth` action type that takes in a detected mouth center. Rather, the `MoveToMouth` action can use the generic `MoveTo` action type.

See [feeding_web_interface#63](https://github.com/personalrobotics/feeding_web_interface/pull/63) for how this was tested.


Raidakarim (approved):
LGTM

PR #29, opened 2023-06-13
Update README.md re. Launching RealSense node on nano

amalnanavati:
Added instructions on launching the RealSense nodes on nano. I've verified that this works on t0b1.


egordon (approved):
lgtm

PR #36, opened 2023-07-14
Implement Face Detection

### LINKED ISSUE #27: [ros2-devel] Face Detection
The ROS2 ADA feeding system needs face detection. This issue encompasses the following:

1. Read and understand the [ROS1 face detection code](https://github.com/personalrobotics/face_detection/tree/master) (in C++) we used.
2. Familiarize yourself with the desired interface for ROS2 face detection, which consists of a ROS service to toggle face detection on and off, and while it is on it publishes results to a ROS topic. The best way to familarize yourself with this is to look at the [face detection dummy node](https://github.com/personalrobotics/feeding_web_interface/blob/main/feeding_web_app_ros2_test/feeding_web_app_ros2_test/FaceDetection.py).
3. Investigate ways to do face detection in Python. We not only need to detect the stomion (mouth center) in the 2D image, but we also have to translate it into a 3D point relative to the camera frame.
4. Implement face detection in Python.

taylorkf:
# Description

This PR adds a face detection node, addressing https://github.com/personalrobotics/ada_feeding/issues/27. It also addresses some issues with segment_from_point (python 3 typing errors and a function call that was possibly blocking realsense nodes).

# Testing procedure

On the realsense camera (using t0b1/nano), I tested that the node starts and stops with the appropriate service call, and publishes mouth locations only when a face is detected.

**Initialization**:
1. SSH into nano@nano
2. On nano, start vision nodes:
    1. `ros2config`
    2. `ros2 launch realsense2_camera rs_launch.py rgb_camera.profile:='640,480,30' depth_module.profile:='640,480,30' align_depth.enable:='true' initial_reset:='true'`
3. On t0b1, launch the face_detection node:
    1. `ros2 launch ada_feeding_perception ada_feeding_perception_launch.xml`
4. Listen to the /face_detection topic:
    1. `ros2 topic echo /face_detection`

**Testing**:
1. For each of the following commands, run it and verify that it executes as expected. 
    1. Turn on face detection 
        - [x] `ros2 service call /toggle_face_detection std_srvs/srv/SetBool "{data: True}"`
        - [x] The face detection node should receive this message
   2. Turn off face detection
        - [x]  `ros2 service call /toggle_face_detection std_srvs/srv/SetBool "{data: False}"`
        - [x] The face detection node should receive this message
2. Position your face in front of the camera and move to several different positions. Make sure that the face_detection topic returns the correct values.
     

# Before Merging
- [ ] `Squash & Merge`


ada_feeding_perception/package.xml
```diff
@@ -11,6 +11,7 @@
   <test_depend>ament_flake8</test_depend>
   <test_depend>ament_pep257</test_depend>
   <test_depend>python3-pytest</test_depend>
+  <test_depend>opencv-python</test_depend>
```
amalnanavati, line 14: Fascinating that this had to be a `test_depend` and nothing else. 

No need to change this, I just find it interesting

ada_feeding_perception/ada_feeding_perception/face_detection.py
```diff
@@ -0,0 +1,314 @@
...
+                    # Add a dummy face marker to the image
+                    cv_image = self.bridge.imgmsg_to_cv2(msg, "bgr8")
+                    for face in faces:
+                        # Draw a white coloured rectangle around each face
+                        (x, y, w, d) = face
+                        cv2.rectangle(
+                            cv_image, (x, y), (x + w, y + d), (255, 255, 255), 2
+                        )
+
+                # Display mouth landmarks (landmarks 48-67 are mouth landmarks, and the length of landmark[0] is 68)
```
amalnanavati, line 285: Is there a website you can link to that says what each landmark corresponds to? That would be helpful for future folks to understand this code.

ada_feeding_perception/ada_feeding_perception/face_detection.py
```diff
@@ -0,0 +1,314 @@
...
+                self.is_face_detected = True
+
+            if self.is_face_detected:
+                # Create an instance of the Facial landmark Detector with the model
+                landmark_detector = cv2.face.createFacemarkLBF()
+                landmark_detector.loadModel(self.landmark_model_path)
+
+                # Detect landmarks on image
+                _, landmarks = landmark_detector.fit(image_gray, faces)
+                for landmark in landmarks:
```
amalnanavati, line 274: I'm really confused by this loop over `landmarks`. You passed into `landmark_detector.fit` a list of faces. So presumably landmarks is a 2D list where the first dimension corresponds to faces, and the second dimension corresponds to the landmarks within that face, right?

Then, why are you looping over landmarks, and then looping over faces? Isn't that essentially looping twice over faces? You also don't refer to the variable `landmark`. Further, it seems like the main functional part of this loop is annotating `cv_image`. But (1) you are re-generating that variable for every landmark; and (b) you are re-creating a variable you already created above.

I'm suggest the following:
1. At the top of the function, create two CV images, one to annotate and one grayscale to pass into the detectors.
2. Move the `for face in faces` logic after faces have been detected. It doesn't seem like it actually has anything to do with the landmarks.

ada_feeding_perception/ada_feeding_perception/face_detection.py
```diff
@@ -0,0 +1,314 @@
...
+                    cv_image = self.bridge.imgmsg_to_cv2(msg, "bgr8")
+                    for face in faces:
+                        # Draw a white coloured rectangle around each face
+                        (x, y, w, d) = face
+                        cv2.rectangle(
+                            cv_image, (x, y), (x + w, y + d), (255, 255, 255), 2
+                        )
+
+                # Display mouth landmarks (landmarks 48-67 are mouth landmarks, and the length of landmark[0] is 68)
+                for i in range(48, len(landmark[0])):
```
amalnanavati, line 286: You refer to `landmark` here, but this is outside the loop `for landmark in landmarks` which is confusing and seems rife for a bug.

ada_feeding_perception/ada_feeding_perception/face_detection.py
```diff
@@ -0,0 +1,314 @@
...
+            if len(faces) > 0:
+                self.is_face_detected = True
+
+            if self.is_face_detected:
+                # Create an instance of the Facial landmark Detector with the model
+                landmark_detector = cv2.face.createFacemarkLBF()
+                landmark_detector.loadModel(self.landmark_model_path)
+
+                # Detect landmarks on image
+                _, landmarks = landmark_detector.fit(image_gray, faces)
```
amalnanavati, line 273: Can you add a comment here explaining the structure of `landmarks` (e.g., is it a 2D list, a 3D list, etc. and what does each dimension represent?)

ada_feeding_perception/ada_feeding_perception/face_detection.py
```diff
@@ -0,0 +1,314 @@
...
+                            cv_image, (x, y), (x + w, y + d), (255, 255, 255), 2
+                        )
+
+                # Display mouth landmarks (landmarks 48-67 are mouth landmarks, and the length of landmark[0] is 68)
+                for i in range(48, len(landmark[0])):
+                    x, y = landmark[0][i]
+                    cv2.circle(cv_image, (int(x), int(y)), 1, (0, 255, 0), 5)
+
+                annotated_msg = self.bridge.cv2_to_imgmsg(cv_image, "bgr8")
+                annotated_img = annotated_msg
```
amalnanavati, line 291: `annotated_img` actually seems like a misnamed variable since you have been making the distinction between `images` and `msgs`.

Why not remove the line `annotated_img = annotated_msg` and then change like 297 (annotated_img = msg) to `annotated_msg = msg` and then publish `annotated_msg`? That seems like a clearer variable name, and then prevents creating two variables (annotated_msg and annotated_img) that contain the same data.

ada_feeding_perception/ada_feeding_perception/face_detection.py
```diff
@@ -0,0 +1,314 @@
...
+from rclpy.parameter import Parameter
+from rcl_interfaces.msg import ParameterDescriptor, ParameterType
+
+
+from ada_feeding_perception.helpers import (
+    download_checkpoint,
+)
+
+
+class FaceDetectionNode(Node):
```
amalnanavati, line 28: Technically, for best coding practices you should have a class docstring as well. Can you add one below the class definition? Feel free to have it be mostly the same as your module docstring at the top of the file.

ada_feeding_perception/ada_feeding_perception/face_detection.py
```diff
@@ -0,0 +1,314 @@
...
+            ToggleFaceDetection,
+            "ToggleFaceDetection",
+            self.toggle_face_detection_callback,
+        )
+
+        # Subscribe to the camera feed
+        self.img_subscription = self.create_subscription(
+            Image, "camera/color/image_raw", self.camera_callback, 1
+        )
+        self.img_subscription  # prevent unused variable warning
```
amalnanavati, line 103: I don't think this is necessary (neither is the one below). Because they are class attributes (e.g., `self.`) the linter won't give an unused variable warning here.

ada_feeding_perception/ada_feeding_perception/face_detection.py
```diff
@@ -0,0 +1,314 @@
...
+                        name="model_dir",
+                        type=ParameterType.PARAMETER_STRING,
+                        description="The location of the directory where the model checkpoint is / should be stored",
+                        read_only=True,
+                    ),
+                ),
+            ],
+        )
+
+    def toggle_face_detection_callback(self, request, response):
```
amalnanavati, line 187: Please add type annotations here. I think it should be:

`def toggle_face_detection_callback(self, request: ToggleFaceDetection.Request, response: ToggleFaceDetection.Response) -> ToggleFaceDetection.Response:`

ada_feeding_perception/ada_feeding_perception/face_detection.py
```diff
@@ -0,0 +1,314 @@
...
+            self.is_on_lock.release()
+            response.face_detection_is_on = True
+        else:
+            self.is_on_lock.acquire()
+            self.is_on = False
+            self.is_on_lock.release()
+            response.face_detection_is_on = False
+        return response
+
+    def depth_callback(self, msg):
```
amalnanavati, line 208: Type annotations

ada_feeding_perception/ada_feeding_perception/face_detection.py
```diff
@@ -0,0 +1,314 @@
...
+            mouth_location.point.y = float(v)
+            mouth_location.point.z = float(depth)
+
+            # Publish 3d point
+            face_detection_msg = FaceDetection()
+            face_detection_msg.is_face_detected = self.is_face_detected
+            face_detection_msg.detected_mouth_center = mouth_location
+            self.publisher_results.publish(face_detection_msg)
+
+    def camera_callback(self, msg):
```
amalnanavati, line 241: Type annotations

ada_feeding_perception/package.xml
amalnanavati, line 1: Could you add the multiple maintainer's emails to this file as well, as you did in `setup.py`?

ada_feeding_perception/launch/ada_feeding_perception_launch.xml
```diff
@@ -1,7 +1,7 @@
 <launch>
-  <node pkg="ada_feeding_perception" exec="segment_from_point" name="segment_from_point">
-    <param from="$(find-pkg-share ada_feeding_perception)/config/segment_from_point.yaml"/>
+  <!-- Launch face_detection -->
```
amalnanavati, line 2: I don't like the idea of pushing a change where we are removing food detection. I think it was fine for testing face detection, but we should identify the issue before merging the changes in, so that face detection and food detection can both run at the same time.

ada_feeding_perception/ada_feeding_perception/face_detection.py
```diff
@@ -0,0 +1,314 @@
...
+            # Retrieve the depth value for this 2d coordinate
+            width = msg.width
+            depth = msg.data[(u % width) + (v * width)]
+
+            # Create target 3d point
+            mouth_location = PointStamped()
+            mouth_location.header = msg.header
+            mouth_location.point.x = float(u)
+            mouth_location.point.y = float(v)
+            mouth_location.point.z = float(depth)
```
amalnanavati, line 233: The depth image coordinates are in `mm`. However, the norm in ROS is for `PoseStamped` points to be in `m`. Can you convert them before putting them into the message?

amalnanavati (changes requested):
See in-line comments. There is one major bug re. converting 2D pixels to 2D points, some code that seems unnecessary / doesn't make sense regarding iterating over landmarks, and most of the rest of the in-line comments are small.

Additionally, I have four broader points of discussion:
1. It seems like your code finds the stomion of the mouth in the RGB image and then gets its depth in the depth image. But if the mouth is open, wouldn't the depth of the stomion be really far into the mouth/throat? I think we should use the (u,v) position of the stomion, but for the depth we should average (or min or max) all the detected landmarks along the lips.
2. After detecting a mouth center, it looks like you get the depth of the mouth on the *next* depth image (as opposed to on the *closest* one). In practice, at what rate do depth images come in relative to the RGB images; is it actually the case that the next one will be the closest one? Because if we are using face detection during visual servoing (i.e., while the robot arm is moving) then even a 0.25 sec delay between those images could result in the (u,v) coordinates that were the mouth center now being on the nose, or to the side of the face, resulting in wrong depth measurements.
    1. To address this, we might need to store depth images e.g., from the last second, and find the closest depth image to the RGB one, not just the next depth image.
3. What if multiple faces are detected? (e.g., Tyler's caregiver is standing behind Tyler, or a fascinated friend is staring at the robot). Firstly, what would the detector do in that case? Secondly, I think what it *should do* is have an "expected face location" (e.g., near wheelchair headrest) in a fixed world frame that is passed in as a parameter, and it selects the face that is closest to that point.
4. I'm concerned that our testing procedure is not rigorous enough, given that it overlooked the bug re. the conversion from pixels to 3D coordinates. Can you add an additional test, where you run the face detector, move your face around, and ensure that the detected point actually aligns with your real mouth's distance from the camera? Try your head in several different poses, and the robot arm in several different poses.

taylorkf:
@amalnanavati Thanks for the bug catches! My lack of experience working with cv is showing 😆 Actually writing down a quick algorithm this time and will get this coded up on Monday when I'm back in town.

ada_feeding_perception/ada_feeding_perception/face_detection.py
```diff
@@ -0,0 +1,314 @@
...
+        self.is_on_lock = Lock()
+
+        # Keeps track of the detected mouth center
+        self.img_mouth_center = []
+        self.is_face_detected = False
+
+        # Create the service
+        self.srv = self.create_service(
+            ToggleFaceDetection,
+            "ToggleFaceDetection",
```
amalnanavati, line 95:
```suggestion
            "toggle_face_detection",
```
[ROS naming conventions](http://wiki.ros.org/ROS/Patterns/Conventions) have topics and services be in snake_case. (I know this was a carryover from the dummy node -- that was my mistake)

ada_feeding_perception/ada_feeding_perception/face_detection.py
```diff
@@ -0,0 +1,388 @@
...
+    download_checkpoint,
+    get_img_msg_type,
+    cv2_image_to_ros_msg,
+)
+
+
+class FaceDetectionNode(Node):
+    """
+    This node publishes a 3d PointStamped location
+    of the largest detected mouth with respect to camera_depth_optical_frame.
```
amalnanavati: nit: Largest detected _mouth_ or _face_?

ada_feeding_perception/ada_feeding_perception/face_detection.py
```diff
@@ -0,0 +1,388 @@
...
+    """
```
amalnanavati, line 41: In line with our conversation with Ethan, can you add a TODO here that eventually this node should return all detected faces (perhaps with an ID if this node is tracking faces) and then let the client decide which to use? (As opposed to what currently happens where this node decides which of multiple faces to return, based on the size in pixel space.)

ada_feeding_perception/ada_feeding_perception/face_detection.py
```diff
@@ -0,0 +1,388 @@
...
+        # is initialized will not be reflected.
+        (
+            face_model_name,
+            face_model_base_url,
+            landmark_model_name,
+            landmark_model_base_url,
+            model_dir,
+        ) = self.read_params()
+        self.face_model_name = face_model_name.value
+        self.landmark_model_name = landmark_model_name.value
```
amalnanavati: It looks like both `face_model_name` and `landmark_model_name` are only used in the `__init__` function and therefore don't need to be class attributes. If that is truly the case, can you just make them local variables instead?

ada_feeding_perception/ada_feeding_perception/face_detection.py
```diff
@@ -0,0 +1,388 @@
...
+                self.landmark_model_name, model_dir.value, landmark_model_base_url.value
+            )
+            self.get_logger().info(
+                f"Model checkpoint downloaded {self.landmark_model_path}."
+            )
+
+        # Keeps track of whether face detection is on or not
+
+        self.is_on = False
+        self.is_on_lock = Lock()
```
amalnanavati, line 98: Nit: Can you initialize these with the service that changes these values (i.e., on line 117) to help with readability?

ada_feeding_perception/ada_feeding_perception/face_detection.py
```diff
@@ -0,0 +1,388 @@
...
+        self.bridge = CvBridge()
+
+        self.is_face_detected = False
+        self.depth_buffer = collections.deque(maxlen=10)
```
amalnanavati: Same here -- if you could define the depth buffer along with the depth subscriber (before, but still in adjacent lines of code) that would help me intuit what its purpose is. Currently, I had to Ctrl+F and look further in the code to understand what this variable's purpose is.

ada_feeding_perception/ada_feeding_perception/face_detection.py
```diff
@@ -0,0 +1,388 @@
...
+        self.img_timestamp = None
```
amalnanavati: Which img is this? The latest img received or something else? The color image or depth image? At a minimum a comment would help here, but also moving this line to be near the subscriber it is used for will help with readability.

ada_feeding_perception/ada_feeding_perception/face_detection.py
```diff
@@ -0,0 +1,388 @@
...
+
+        # Approximate values of current camera info (updated with subscription)
```
amalnanavati: Can you specify _which_ values these are? It looks like it is the intrinsics camera matrix (the [K matrix in CameraInfo](http://docs.ros.org/en/melodic/api/sensor_msgs/html/msg/CameraInfo.html))

ada_feeding_perception/ada_feeding_perception/face_detection.py
```diff
@@ -0,0 +1,388 @@
...
+        self, request: SetBool.Request, response: SetBool.Response
+    ) -> SetBool.Response:
+        """
+        Callback function for the toggle_face_detection service. Safely toggles
+        the face detection on or off depending on the request.
+        """
+        self.get_logger().info(f"Incoming service request. data: {request.data}")
+        with self.is_on_lock:
+            self.is_on = request.data
+        response.success = request.data
```
amalnanavati: The way I [use this in `MoveToMouth`](https://github.com/personalrobotics/ada_feeding/blob/a6a6655cfb5b6a05a91a120b906175be2cb9fcb7/ada_feeding/ada_feeding/trees/move_to_mouth_tree.py#L149), I expect the result to indicate whether toggling succeeded or not. In other words, if the request was `False` and you successfully turned it off, the response should be `{'success': True}`. Also, can you also populate [the `message` field of SetBool](http://docs.ros.org/en/noetic/api/std_srvs/html/srv/SetBool.html)?

I think the best way to do this is:
```
response.success = False
response.message = f"Failed to set is_on to {request.data}"
with self.is_on_lock:
    self.is_on = request.data
    response.success = True
    response.message = f"Successfully set is_on to {request.data}"
return response
```

ada_feeding_perception/ada_feeding_perception/face_detection.py
```diff
@@ -0,0 +1,388 @@
...
+        return response
+
+    def camera_info_callback(self, msg: CameraInfo):
+        """
+        Callback function for the toggle_face_detection service. Safely toggles
+        the face detection on or off depending on the request.
+        """
+        self.camera_matrix = msg.k
```
amalnanavati, line 276: Since `self.camera_matrix` can be accessed across multiple threads, it should have a lock. It looks like lines 294-298 in `depth_callback` below use it. I'd suggest creating a local copy of the variable at the top of `depth_callback`, like you do with `self.is_on`/`self.is_on_lock`, and then using that local copy (not the instance attribute attribute) throughout the function

ada_feeding_perception/ada_feeding_perception/face_detection.py
```diff
@@ -0,0 +1,388 @@
...
+        with self.is_on_lock:
+            self.is_on = request.data
+        response.success = request.data
+        return response
+
+    def camera_info_callback(self, msg: CameraInfo):
+        """
+        Callback function for the toggle_face_detection service. Safely toggles
+        the face detection on or off depending on the request.
+        """
```
amalnanavati, line 275: Can you add the following in this docstring?

```
TODO: We technically only need to read one message from CameraInfo, since the intrinsics don't change. If `rclpy` adds a `wait_for_message` function, this subscription/callback should be replaced with that.
```

ada_feeding_perception/ada_feeding_perception/face_detection.py
amalnanavati: Also, can you make the length of `self.depth_buffer` a parameter with default value 10? Because if in the future we increase the FPS of the depth camera, the number of images stored in the buffer will also need to be increased proportionally.

ada_feeding_perception/ada_feeding_perception/face_detection.py
```diff
@@ -0,0 +1,388 @@
...
+
+        # Collect last ten depth images (depth_buffer is a collections.deque of max length 10)
+        self.depth_buffer.append(msg)
+        with self.is_on_lock:
+            is_on = self.is_on
+
+        # Check if face detection is on
+        if is_on:
+            face_detection_msg = FaceDetection()
+            if self.is_face_detected:
```
amalnanavati: I think `self.is_face_detected` should also be protected by a lock, since it can be accessed across multiple threads. Same with the other attributes that are set in `color_callback` and used here: `self.img_mouth_center`, `self.img_timestamp`, and `self.img_mouth_points`. The main reason for this is to ensure that all these attributes are read from a message of the same timestamp. As it stands, if a call to `color_callback` is being run at the same time as this `depth_callback`, you might get the `img_mouth_center` from the previous call but the `img_mouth_points` from the new call, which is undesirable.

amalnanavati: To address this, I'd suggest having one lock `self.mouth_detection_lock` to handle all instance attributes that are set in `color_callback`. In `color_callback`, do all your computations with local variables, and only at the end of the function you should acquire the lock and set all the instance attributes. Similarly, at the top of `depth_callback` you should acquire the lock, store all instance attributes as local variables, and then only use the local variables for the rest of the function.

ada_feeding_perception/ada_feeding_perception/face_detection.py
```diff
@@ -0,0 +1,388 @@
...
+                # Retrieve the 2d location of the mouth center
+                u = int(self.img_mouth_center[0])
+                v = int(self.img_mouth_center[1])
+
+                # Find depth image closest in time to saved color image
+                difference_array = np.array([])
```
amalnanavati: Nit: since you know the length of `difference_array` before-hand (the same length as `depth_img`) you should instead create an array of that size (e.g., `difference_array = np.zeros((len(self.depth_buffer),), stype=np.float32)`) and then iterate over the indices (`for i in range(len(self.depth_buffer)`) and set that index of the difference array instead of appending (`difference_array[i] = abs(depth_time - img_time)`). This prevents multiple changes of the size of the underlying array.

ada_feeding_perception/ada_feeding_perception/face_detection.py
```diff
@@ -0,0 +1,388 @@
...
+                # Retrieve the depth value averaged over all mouth coordinates
+                depth_sum = 0
+                for point in self.img_mouth_points:
+                    depth_sum += closest_depth[int(point[1])][int(point[0])]
+                depth = depth_sum / float(len(self.img_mouth_points))
+
+                # Create target 3d point, with mm measurements converted to m
+                mouth_location = PointStamped()
+                mouth_location.header = msg.header
+                mouth_location.point.x = (
```
amalnanavati, line 336: Can you add a link to the YouTube video you got these formulae from, so a reader of the code can refer back to that if need be?

ada_feeding_perception/ada_feeding_perception/face_detection.py
```diff
@@ -0,0 +1,388 @@
...
+
+    def camera_callback(self, msg: CompressedImage):
+        """
+        Callback function for the camera feed. If face detection is on, this
+        function will detect faces in the image and store the location of
+        the largest detected face/mouth.
+        """
+        # pylint: disable=too-many-locals
+        # Two variables over the limit is fine, needed to find both faces and mouths
+        image_rgb = cv2.imdecode(np.frombuffer(msg.data, np.uint8), cv2.IMREAD_COLOR)
```
amalnanavati: Looks like this line isn't used unless `is_on` is True -- therefore, can you move it to inside that conditional?

ada_feeding_perception/ada_feeding_perception/face_detection.py
```diff
@@ -0,0 +1,388 @@
...
+        with self.is_on_lock:
+            is_on = self.is_on
+
+        if is_on:
+            self.is_face_detected = False
+
+            # Convert image to Grayscale
+            image_gray = cv2.cvtColor(image_rgb, cv2.COLOR_RGB2GRAY)
```
amalnanavati: JW, did you do an `imshow` after loading `image_rgb` to check whether it is actually RGB or BGR? OpenCV's default way to store images is BGR whereas ROS's is RGB. You should do:

```
cv2.imshow("image_rgb", image_rgb)
cv2.waitKey(0)
```

If that image's colors appear correctly, that means it is BGR, which means the above line should be `cv2.COLOR_BGR2GRAY`. If that image's colors appear incorrectly, that means it is RGB, which means the above line is correct.

ada_feeding_perception/ada_feeding_perception/face_detection.py
```diff
@@ -0,0 +1,388 @@
...
+
+                # Retrieve the depth value averaged over all mouth coordinates
+                depth_sum = 0
+                for point in self.img_mouth_points:
+                    depth_sum += closest_depth[int(point[1])][int(point[0])]
+                depth = depth_sum / float(len(self.img_mouth_points))
+
+                # Create target 3d point, with mm measurements converted to m
+                mouth_location = PointStamped()
+                mouth_location.header = msg.header
```
amalnanavati: I'd actually prefer if the header is the image message header as opposed to the depth message header, so that there is a one-to-one correspondence between the headers in the annotated image message and the FaceDetection message.  Since you already have the img_timestamp, it is just a matter of `mouth_location.header = Header(stamp=img_timestamp)`

ada_feeding_perception/ada_feeding_perception/face_detection.py
```diff
@@ -0,0 +1,388 @@
...
+            is_on = self.is_on
+
+        if is_on:
+            self.is_face_detected = False
+
+            # Convert image to Grayscale
+            image_gray = cv2.cvtColor(image_rgb, cv2.COLOR_RGB2GRAY)
+
+            # Create an instance of the Face Detection Cascade Classifier
+            detector = cv2.CascadeClassifier(self.face_model_path)
```
amalnanavati: Just wondering, does this need to be re-created for every image message, or can you create it once in `__init__` and reuse it for every image message? (The latter is more efficient if it works.)

ada_feeding_perception/ada_feeding_perception/face_detection.py
```diff
@@ -0,0 +1,388 @@
...
+            faces = detector.detectMultiScale(image_gray)
+
+            # Check if a face has been detected
+            if len(faces) > 0:
+                self.is_face_detected = True
+
+            if self.is_face_detected:
+                # Create an instance of the Facial landmark Detector with the model
+                landmark_detector = cv2.face.createFacemarkLBF()
+                landmark_detector.loadModel(self.landmark_model_path)
```
amalnanavati: Also wondering if this can be initialized/loaded just once in `__init__` instead of every time in the callback. Loading the models probably slows the callback down quite a bit.

ada_feeding_perception/ada_feeding_perception/face_detection.py
```diff
@@ -0,0 +1,388 @@
...
+
+                # Detect landmarks (a 3d list) on image
+                # Relevant dimensions are 1: faces and 3: individual landmarks 0-68
+                _, landmarks = landmark_detector.fit(image_gray, faces)
+
+                # Add face markers to the image and find largest face
+                largest_face = [0, 0]
```
amalnanavati:
```suggestion
                largest_face = [0, 0] # [area (px^2), index]
```

ada_feeding_perception/ada_feeding_perception/face_detection.py
```diff
@@ -0,0 +1,388 @@
...
+                for i, face in enumerate(faces):
+                    # Draw a white rectangle around each face
+                    (x, y, w, h) = face
+                    cv2.rectangle(image_rgb, (x, y), (x + w, y + h), (255, 255, 255), 2)
+
+                    largest_face = [max(largest_face[0], w * h), i]
+                    # Display mouth landmarks (48-67, as explained in the below link)
+                    # https://pyimagesearch.com/2017/04/03/facial-landmarks-dlib-opencv-python/
+                    for j in range(48, 67):
```
amalnanavati:
```suggestion
                    for j in range(48, 68):
```

Python is exclusive of the upper bound.

ada_feeding_perception/ada_feeding_perception/face_detection.py
```diff
@@ -0,0 +1,388 @@
...
+                        x, y = landmarks[i][0][j]
+                        cv2.circle(image_rgb, (int(x), int(y)), 1, (0, 255, 0), 5)
+
+                # Draw a red rectangle around largest face
+                (x, y, w, h) = faces[largest_face[1]]
+                cv2.rectangle(image_rgb, (x, y), (x + w, y + h), (255, 0, 0), 2)
+
+                # Find stomion (mouth center) in image
+                self.img_mouth_center = landmarks[largest_face[1]][0][66]
+                self.img_mouth_points = landmarks[largest_face[1]][0][48:67]
```
amalnanavati:
```suggestion
                self.img_mouth_points = landmarks[largest_face[1]][0][48:68]
```

ada_feeding_perception/ada_feeding_perception/helpers.py
```diff
@@ -185,8 +186,11 @@ def bbox_from_mask(mask: np.ndarray[bool]) -> BoundingBox:
 
 
 def crop_image_mask_and_point(
-    image: np.ndarray, mask: np.ndarray[bool], point: Tuple[int, int], bbox: BoundingBox
-) -> Tuple[np.ndarray, np.ndarray[bool], Tuple[int, int]]:
+    image: np.ndarray,
+    mask: npt.NDArray[bool],
+    point: Tuple[int, int],
+    bbox: BoundingBox,
+) -> Tuple[np.ndarray, npt.NDArray[bool], Tuple[int, int]]:
```
amalnanavati: I'm surprised that this `np.ndarray` didn't cause an issue. Anyway, no need to change it as it is getting fixed in #47

ada_feeding_perception/config/face_detection.yaml
```diff
@@ -0,0 +1,14 @@
...
+    face_model_name: haarcascade_frontalface_alt2.xml
+    # The URL to download the face detection model checkpoint from if it is not already downloaded
+    face_model_base_url: "https://raw.githubusercontent.com/opencv/opencv/master/data/haarcascades/"
+    # The name of the facial landmark detection model checkpoint to use
+    landmark_model_name: lbfmodel.yaml
+    # The URL to download the facial landmark detection model checkpoint from if it is not already downloaded
+    landmark_model_base_url: "https://raw.githubusercontent.com/kurnianggoro/GSOC2017/master/data/"
+
+
+
```
amalnanavati: Nit: I think convention is to have just a single newline at the bottom of every file (not multiple).

amalnanavati (changes requested):
Overall it looks good! Most of the below comments are nits to help with readability. The only functional changes are:
1. Changing how `toggle_face_detection` response is handled.
2. Adding locks to make this node thread-safe.
3. (If possible) move model loading to `__init__`.
4. Fixing indexing for mouth landmarks.

ada_feeding_perception/ada_feeding_perception/face_detection.py
amalnanavati: In general, variables that are used in similar ways should be initialized on nearby lines of code.

ada_feeding_perception/ada_feeding_perception/face_detection.py
taylorkf: Since I'm not in to check this today, I'm changing this to BGR2GRAY. Can you check this? Or I can check it once I'm not sick.

amalnanavati:
Tested it, made some slight changes but other than that it looks good.

I'm merging it in, but as a side-note I think the fact that camera subscriptions is unreliable will be a big problem for us. For example, if while running `ada_feeding_perception_launch.xml`, I also use the commond-line tool to check the hz of the aligned depth topic, that is one subscription too many and the face detection node stops receiving color images. We do need an in-depth investigation into when the problem arises, and whether a republisher (started in #49 ) can address it, and if so what the minimal set of topics to republish is.

amalnanavati (approved):

ada_feeding_perception/ada_feeding_perception/face_detection.py
amalnanavati: Verified that BGR2GRAT is correct

PR #35, opened 2023-07-12
Implement py_trees for MoveToConfiguration and MoveToPose

amalnanavati:
# Description

This PR, in service of issue #28, does the following:
1. Implements a generic `MoveTo` behavior that plans and executes a movement using pymoveit2.
2. Implements decorators on top of that behavior to specify goal constraints. This includes `SetJointGoalConstraint`, `SetPositionGoalConstraint`, and `SetOrientationGoalConstraint`
3. Implements a `MoveToConfigurationTree` py_tree that moves the robot arm to a specific configuration.
4. Implements `MoveAbovePlate`, `MoveToRestingPosition`, and `MoveToStowLocation` as instances of the `MoveToConfigurationTree` pytree.
5. Implements a `MoveToPoseTree` py_tree that moves the robot arm to a specific pose.
6. Implements a dummy `MoveToMouth` action that moves the robot to the staging location using `MoveToPoseTree`. This is mainly meant to test `MoveToPoseTree` and will be replaced with the real MoveToMouth later.

## Design Decisions
Recall that all robot motion actions in ada_feeding have the same return types for results and feedback (see [MoveTo.action](https://github.com/personalrobotics/ada_feeding/blob/ros2-devel/ada_feeding_msgs/action/MoveTo.action). With that in mind, we make the following design decisions.
- **Behavior**: `MoveTo` is the behavior for any robot motion. It plans the motion, executes the motion, populates the blackboard with feedback, and handles termination. However, note that it **_does not_** set any goal for planning.
    - Because it sets feedback, this behavior must have access to the parent tree's namespace. Therefore, it has two concurrent blackboards; one for its own variables and another to set feedback variables in the tree's namespace.
- **Decorator**: **_All goal and path constraints are implemented as decorators on top of the `MoveTo` behavior_**. These decorators can be chained together in order to set multiple constraints for a single motion. The shared logic is implemented in `MoveToConstraint`, which all constraint decorators must inherit from.
- **Tree**: `MoveToTree` contains code to initialize variables for feedback on the blackboard, send the goal, get feedback, and get the result. Any py_tree that carries out robot motion should inherit from `MoveToTree` (e.g., `MoveToConfigurationTree`, `MoveToPoseTree`)

# Testing procedure

On the real robot (using lovelace), I tested that the actions executed and canceled, both when I used the ROS command line interface and when I used the web interface.

**Initialization**:
1. Launch MoveIt:
    1. Real Robot: `ros2 launch ada_moveit demo.launch.py`
[...]

PR #40, opened 2023-07-21
Food On Fork Node V2

### LINKED ISSUE #43: [ROS2] Food on Fork Detection
We want a node that can output a probability of whether or not there is Food on Fork.

### LINKED ISSUE #77: Food on Fork Algorithm
For more information, please refer to this [issue](https://github.com/personalrobotics/ada_feeding/issues/43).

atharva-kashyap:
# Description
This PR is adding a feature node Food on Fork that will allow for automatically detecting the presence of food on the fork. It is addressing the issue linked [here](https://github.com/personalrobotics/ada_feeding/issues/43). It will output a probability to a topic `/food_on_fork` and the probability represents the probability with which we are able to predict food on the fork. 

## Design Decisions
* This feature is currently using a model that uses Logistic Regression to make predictions for whether or not there is Food on the Fork.

# Testing procedure
### Connect to Robot
1. `ssh nano` 
2. `ros2config` 
3. `ros2 launch realsense2_camera rs_launch.py rgb_camera.profile:='640,480,30' depth_module.profile:='640,480,30' align_depth.enable:='true' initial_reset:='true'`

### Launching and Testing
1. Make sure to clone this repository and `atharvak/food_on_fork_node_v2` branch and make sure to complete the installation of all dependencies.
2. Inside the `ada_feeding_perception/config` folder, there is a `food_on_fork.yaml` file. This file contains all the parameters. Notice that there is a `test` parameter. Go ahead and change the value of this parameter to be `True`. 
3. Make sure to build and then launch the `food_on_fork` node by running `ros2 launch ada_feeding_perception ada_feeding_perception_launch.xml`. This will launch the node with the specified parameters.
5. In another terminal, please listen to the `/food_on_fork` topic (`ros2 topic echo /food_on_fork`). This is where you will be able to see the predictions being published. 

### Viewing output
* To view the `Prediction Probability`, you can note this either in the log statements or in the terminal where you are listening to the `/food_on_fork` node.
* To view the `num pixels`, you can note this in the log statements.

### Choosing a fork and food item
* Please choose a fork with fork tines NOT covered in tape (so, regular fork)
* Please choose any food item for testing (You could consider doing a solid and a leafy food item, if you wish)

Figure 1: Logistic Regression Graph

## Test 1: Verify the overall behavior/working of the node (position: `Home`)
Position the robot to be in the `Home` position. Then, launch the node and make sure to echo the `/food_on_fork` topic. 
1. [ ] Make sure you are seeing some log statements that are logging `num pixels` and `Prediction Probability` (if this is not working, make sure to go back and follow [Launching and Testing steps](#-Launching-and-Testing))
2. [ ] Verify that when you are listening to topic `/food_on_fork`, where you see a probability (0.0-1.0) being output

## Test 2: Verify the accuracy of count of pixels and output probability (position: `Home`; no Food on Fork)
Position the robot to be in the `Home` position. Then, launch the node and make sure to echo the `/food_on_fork` topic. Do not add any food onto the fork. 
1. [ ] Verify that the `num pixels` is in the range of 0-150 
2. [ ] Based on the number of pixels, visually trace over to a location of those number of pixels on the x-axis of the above figure. Then, visually verify that the `Prediction Probability` output is in a reasonable range on the y-axis of the above figure. 

## Test 3: Verify the accuracy of count of pixels and output probability (position: `Home`; Add Food on Fork)
* Position the robot to be in the `Home` position. Then, launch the node and make sure to echo the `/food_on_fork` topic.
* Add a food item on to the fork
1. [ ] Verify that the `num pixels` is in a higher range than when there was no food on the fork.
2. [ ] Based on the number of pixels, visually trace over to a location of those number of pixels on the x-axis of the above figure. Then, visually verify that the `Prediction Probability` output is in a reasonable range on the y-axis of the above figure. 

## Test 4: Repeat Tests 2-3 with a change in position to `In-Front of Face`
Position the robot to be in the `In-Front of Face` position. Make sure that the fork is approximately 2 inches away from the mannequin's mouth. Then, launch the node and make sure to echo the `/food_on_fork` topic. Then repeat Tests 2-3 in this position.

### When there is no Food on Fork
1. [ ] Verify that the `num pixels` is in the range of 0-150 
2. [ ] Based on the number of pixels, visually trace over to a location of those number of pixels on the x-axis of the above figure. Then, visually verify that the `Prediction Probability` output is in a reasonable range on the y-axis of the above figure. 

### When there is Food on Fork
1. [ ] Verify that the `num pixels` is in a higher range than when there was no food on the fork.
2. [ ] Based on the number of pixels, visually trace over to a location of those number of pixels on the x-axis of the above figure. Then, visually verify that the `Prediction Probability` output is in a reasonable range on the y-axis of the above figure. 

## Test 5: Repeat Tests 2-3 with a change in position to `Over-the-plate`
Position the robot to be in the `Over-the-plate` position. Make sure that the fork is approximately 2 inches away from plate. Then, launch the node and make sure to echo the `/food_on_fork` topic. Then repeat Tests 2-3 in this position.

### When there is no Food on Fork
1. [ ] Verify that the `num pixels` is in the range of 150-300 
2. [ ] Based on the number of pixels, visually trace over to a location of those number of pixels on the x-axis of the above figure. Then, visually verify that the `Prediction Probability` output is in a reasonable range on the y-axis of the above figure. 

### When there is Food on Fork
1. [ ] Verify that the `num pixels` is in a higher range than when there was no food on the fork.
2. [ ] Based on the number of pixels, visually trace over to a location of those number of pixels on the x-axis of the above figure. Then, visually verify that the `Prediction Probability` output is in a reasonable range on the y-axis of the above figure. 

## Test 6: Repeat Tests 2-3 when the arm is In-Motion
* Repeat Tests 2 and 3 by moving the arm around
### When there is no Food on Fork
1. [ ] Verify that the `num pixels` is in the range of 0-300 
2. [ ] Based on the number of pixels, visually trace over to a location of those number of pixels on the x-axis of the above figure. Then, visually verify that the `Prediction Probability` output is in a reasonable range on the y-axis of the above figure. 

### When there is Food on Fork
1. [ ] Verify that the `num pixels` is in a higher range than when there was no food on the fork.
2. [ ] Based on the number of pixels, visually trace over to a location of those number of pixels on the x-axis of the above figure. Then, visually verify that the `Prediction Probability` output is in a reasonable range on the y-axis of the above figure. 

**Note:** After the completion of testing, make sure to go back to the `food_on_fork.yaml` file and change back the test boolean from `True` to `False`. 

- [ ] Run your code through [pylint](https://pylint.readthedocs.io/en/latest/) and address (most) warnings/errors: `pylint --recursive=y .`

# Before Merging
- [ ] `Squash & Merge`


ada_feeding_perception/ada_feeding_perception/food_on_fork.py
```diff
@@ -0,0 +1,136 @@
...
+        if depth_img is not None:
+            num_pixels = self.food_on_fork_num_pixels(depth_img)
+            # print("num pixels: ", num_pixels)
+            prediction = self.predict_food_on_fork(num_pixels)
+            # print("prediction: ", prediction)
+            float32msg = Float32()
+            float32msg.data = prediction
+            self.publisher_depth.publish(float32msg)
+
+    def food_on_fork_num_pixels(self, depth_img,
```
amalnanavati, line 71: In this function and every other function, please add [type hints](https://docs.python.org/3/library/typing.html) for all parameters but `self` and for the return type. See other Python code in this repo as an exmaple.

ada_feeding_perception/ada_feeding_perception/food_on_fork.py
```diff
@@ -0,0 +1,136 @@
...
+                                left_top_corner=(297, 248),
+                                right_bottom_corner=(422, 332),
+                                min_dist=(330 - 20),
+                                max_dist=(330 + 40)) -> int:
```
amalnanavati, line 75: All four of these should be passed in as rosparams instead of function parameters. That way, we can change the parameters without having to modify the code.

ada_feeding_perception/ada_feeding_perception/food_on_fork.py
```diff
@@ -0,0 +1,136 @@
...
+        """
+
+        # consider the points for the rectangle
+        pt1_col, pt1_row = left_top_corner
+        pt2_col, pt2_row = right_bottom_corner
+
+        # create mask that satisfies the rectangle and distance conditions
+        mask_img = np.zeros_like(depth_img, dtype=bool)
+        mask_img[pt1_row:pt2_row, pt1_col:pt2_col] = True
+        mask_img[np.logical_not((min_dist < depth_img) & (depth_img < max_dist))] = False
```
amalnanavati, line 97: I find this logic confusing to follow. Because first you create a mask of all Falses, then you set the entire rectangle to True, and then you set all pixels outside the distance range to False.

I think something like the below code would be more readable (I haven't checked it for syntax errors, but this conveys the general ideal):

```
mask_img = np.zeros_like(depth_img, dtype=bool)
depth_around_fork = depth_img[pt1_row:pt2_row, pt1_col:pt2_col] 
mask_img[pt1_row:pt2_row, pt1_col:pt2_col] = (min_dist < depth_around_fork) & (depth_around_fork < max_dist)
```

ada_feeding_perception/ada_feeding_perception/food_on_fork.py
```diff
@@ -0,0 +1,136 @@
...
+
+        Parameters:
+            num_pixels: number of pixels detected based on which the probability is output
+
+        Returns:
+            probability of the presence of food on fork
+        """
+        # print("num pixels in method: ", num_pixels)
+        num_pixels = np.array([[num_pixels]])
+        num_pixels_reshape = num_pixels.reshape(-1, 1)
```
amalnanavati, line 113: Why do you have to reshape? You already have a 1x1 array even before reshaping

amalnanavati (changes requested):
The code generally looks good. In addition to the in-line comments, I have the following overarching feedback:
1. It looks like you’ve added another dependency, `joblib`. Please document that in the README
2. This node is assuming there is a pre-trained model. However, it is also crucial to have the code you used to train the model, so that: (a) people can understand how you trained it; and (b) future PRL members can retrain it or modify the training if they want. Can you add the training code, as well as the dataset?
3. The model should not be in the `ada_feeding_perception` folder, that should just be Python code. Create a folder `ada_feeding/ada_feeding_perception/model` and put it in there. That also aligns to the folder where Segment Anything downloads its model.
4. Please populate the PR description and testing procedure. The testing procedure should be detailed enough that someone else (e.g., me, Taylor) should be able to run it and recreate your results.
5. Be sure to format your code (see the bottom of the PR description).

atharva-kashyap:
@amalnanavati I have a couple follow-up questions for you from your comments. 

1. What exactly does it mean to add the testing procedure? Would I need to outline the things that someone else needs to do to be able to make sure that the code is working?
2. How/Where should I document dependency? I know you mention readme but would this be overall readme (as linked [here](https://github.com/personalrobotics/ada_feeding/blob/ros2-devel/README.md) but in my branch)?
3. Also, the training code should go into ada_feeding_perception package right? Just as a python file?

amalnanavati:
1. See [this PR](https://github.com/personalrobotics/ada_feeding/pull/35) as an example. You need to:
    1. Come up with a systematic set of tests to run so that you, as well as a critical reviewer, can be convinced that your code works.
    2. You need to document the procedure so that anyone who reads this PR can follow the same procedure and get the same results. Just saying that it passes tests is not enough; it has to be independently verifiable.
2. Add it to the installation section of the `ada_feeding_perception` README. Make the change in this branch (e.g., as part of this PR), and then when this PR get merged into the main branch that change will go along with it.
3. You can put the code in somewhere like `ada_feeding_perception/ada_feeding_perception/food_on_fork_train_model.py`. You can put the data somewhere like `ada_feeding_perception/data/food_on_fork_training_data.csv`. Be sure to add to the README instructions on how to run the model training so that anyone who wants to retrain the model can.

ada_feeding_perception/ada_feeding_perception/food_on_fork.py
atharva-kashyap: @amalnanavati Maybe if I explain my logic, it might be less confusing. I don't mean to downplay your concern but to me those 3 lines makes complete sense. Perhaps if I add my explanation below and into the code, it might make things more clear?
* We start with a mask that is all false. 
```
mask =
[[F, F, F, F],
 [F, F, F, F], 
 [F, F, F, F]]
```
* Then, find the threshold spot using the rectangle and make the mask be true in those spots (indicating that those are the 'voxels' we are considering. For instance, 
```
mask =
[[F, T, T, F],
 [F, T, T, F], 
 [F, T, T, F]]
```
* Then, we use the mask to go in and see which of these voxels are within the range of considered depth and change the mask as follows: 
```
mask =
[[F, F, F, F],
 [F, T, T, F], 
 [F, F, F, F]]
```
* In the end, we just need to count the number of 'voxels' that are true. 

But, I do see your point. In that, in your code, you are simply setting the parts of the mask to true once, rather than making a back-and-forth like I am doing. If you think, computationally/logically, it would be better to refactor and code it up the other way, I would be happy to do so. But, I just wanted to make sure to be on the same page about it :)

ada_feeding_perception/ada_feeding_perception/food_on_fork.py
atharva-kashyap: Yup! You are absolutely right! In the previous version, I had a multiple values since we were trying to train/test to get the metrics. But, when I adapted it to this version, I forgot to change it. I have gone ahead and made that change. Thank you for catching!

ada_feeding_perception/ada_feeding_perception/food_on_fork.py
amalnanavati: Yup I was aware that your code was correct even when I wrote this comment. It's just that:

1. I needed to stare at it for a while to verify that it was correct. IMO the back-and-forth makes it less straightforward for someone else who is reading your code to understand.
2. Computationally, you are setting every F pixel within the ROI twice, whereas I am only doing it once.

TBH I think the bigger benefit of my approach is readability (point 1).

atharva-kashyap: @amalnanavati Personally to me, the original code makes readable sense to me. But, I totally see your point about readability for others. I will make sure to update the code with that! :) Thank you!

amalnanavati: In that case, regardless of which approach you use, add comments (either one longer one at the top, or shorter ones per line) explaining it. That way, even if the reader is unable to understand it with the code alone, they will understand it by reading the comments.

ada_feeding_perception/ada_feeding_perception/food_on_fork_logistic_reg_training.py
```diff
@@ -0,0 +1,50 @@
+import pandas as pd
+import numpy as np
+from sklearn.linear_model import LogisticRegression
+import joblib
+
+
+def main():
+    csv_to_read_from = "../datasets/Ross_7-11-23.csv"
```
amalnanavati, line 8: BTW, why is this file called `Ross`?

ada_feeding_perception/ada_feeding_perception/food_on_fork_logistic_reg_training.py
```diff
@@ -0,0 +1,50 @@
...
+
+    # get the data frame
+    df = pd.read_csv(csv_to_read_from)
+
+    # shuffle all the rows
+    df = df.sample(frac=1)
+
+    # split the dataset (60% train/20% validation/20% test)
+    # Note: the first split occurs at 60%; the second split occurs at 80% and they are all stored in the variables
+    train, validation, test = np.split(df, [int(0.6 * len(df)), int(0.8 * len(df))])
```
amalnanavati, line 18: You never actually use the validation or test sets, right? It seems like you are unnecessarily reducing the size of the training set. 

You should do the following:

1. Run cross-validation, so we actually have a number of how it performs on unseen data, and print that number. See here for an example of [running cross validation](https://scikit-learn.org/stable/modules/cross_validation.html#computing-cross-validated-metrics).
2. Then, after running cross-validation, use the entire set as the train set to re-train the model, and save that. This is better because: (a) it gives you more train data; and (b) it makes the trained model deterministic -- `logistic_reg_model.pkl` will be the exact same every time you run this file.

ada_feeding_perception/config/food_on_fork.yaml
```diff
@@ -0,0 +1,19 @@
...
+    left_top_corner_x: 297
+    # Top-left point's y-val of the bounding box rectangle
+    left_top_corner_y: 248
+    # Bottom-right point's x-val of the bounding box of the rectangle
+    right_bottom_corner_x: 422
+    # Bottom-right point's y-val of the bounding box of the rectangle
+    right_bottom_corner_y: 332
+    # minimum depth to consider (note that 330 is approx distance to the fork tine)
+    min_dist: 310
+    # maximum depth to consider (note that 330 is approx distance to the fork tine)
```
amalnanavati, line 14: Add units to all these comments.

ada_feeding_perception/config/food_on_fork.yaml
```diff
@@ -0,0 +1,19 @@
...
+    max_dist: 370
+    # absolute path for the model
+    model_location: "/home/atharva2/atharvak_node_ws/src/ada_feeding/ada_feeding_perception/model/logistic_reg_model.pkl"
```
amalnanavati, line 17: This should not be an absolute path, it should be a relative path, so it works across computers. See how I handle the model path in Food Segmentation. Fundamentally, I pass in an additional param in the launchfile, `model_dir`, that ROS resolves to be the absolute path to the model directory, and then I append that to the relative path passed into the config file. 

(Note that for that to work, you need to rerun `colcon build` any time you change the model, to get it to actually move into the folder where ROS moves all installed files in the workspace.)

amalnanavati (changes requested):
Code-wise, this looks good. I added a bunch of comments regarding clarity, readability, and style, but the code overall looks close-to-done.

Your test procedure is missing numbers for the probability we should see in the different cases. You basically say “look at the num pixels in the log, and ensure the that the probability outputted is what we would expect when we plug that num pixels into the learned logistic regression function.” This is “obvious” — we know that the node is using a Logistic Regression model with particular parameters, so ofc the output should be the function applied to the input.

Instead, you should come up with an *ideal probability* that you expect to see for *any food-on-fork node* in each case. When there is no food on the fork, should we see a probability between 0-0.1? 0-0.2? Etc. What range of probabilities should we expect to see? When there is food on the fork, what range of probabilities should we see? Add those numbers to the testing procedure — we need target metrics to be able to test the node.

atharva-kashyap:
Closing this PR to give way to PR #104.

PR #39, opened 2023-07-21
Implement Collision Meshes

amalnanavati:
# Description

This PR, building off of #38 , does the following:
1. Added STL messages of the wheelchair, table, head, and wheelchair collision area (expanded from the wheelchair to account for the human's body). All are based on [pr_assets (branch: `amaln/fix_wheelchair_arm`)](https://github.com/personalrobotics/pr_assets/tree/amaln/fix_wheelchair_arm/data/furniture), with some filetype conversions.
2. Added a generic `ada_planning_scene.py` node that takes parameters specifying STL meshes and their poses, and populates the planning scene with those meshes.

Screenshot 2023-07-20 at 8 50 38 PM
Above: An image of the entire planning scene.

Screenshot 2023-07-20 at 8 55 26 PM
Above: An image of the planning scene without the expanded wheelchair collision area.

# Design Decisions
- The way MoveIt works, the planning scene is stored on the MoveIt server, not in individual clients. Therefore, it doesn't make sense to populate the planning scene as a behavior in the behavior tree, since that would unnecessarily repeatedly populate it. Hence, I implemented `ada_planning_scene.py` as a separate node.
- During the course of robot-assisted feeding, we will need to modify the planning scene in at least two instances: (1) updating the head pose based on face detection; and (2) disabling collisions with `wheelchair_collision.stl` (the expanded wheelchair) during the final approach to the mouth. These both can be handled within py_tree behaviors: (1) by [publishing a `MOVE` message to `/collision_object`](https://github.com/ros-planning/moveit_msgs/blob/master/msg/CollisionObject.msg) with the ID of the face; and (2) utilizing ROS2 services to update the allowed_collision_matrix to enable/disable collisions for a particular object (see the [`amaln/allowed_collision_matrix` branch of the PRL fork of pymoveit2](https://github.com/personalrobotics/pymoveit2/tree/amaln/allowed_collision_matrix)). Hence, this node does not need to expose any API for updating the meshes; that can be done by the behaviors directly.

# Testing procedure

I did all the below testing in sim, although I also include initialization instructions to test it on the real robot.

**Initialization**:
1. Launch MoveIt:
    1. Real Robot: `ros2 launch ada_moveit demo.launch.py`
    2. Sim: `ros2 launch ada_moveit demo.launch.py sim:=mock`
2. Launch the F/T sensor:
[...]

PR #42, opened 2023-07-29
Implement MVP MoveToMouth

amalnanavati:
# Description

This PR implements the MVP MoveToMouth. Specifically, it implements a py_tree that:
1. Toggles on face detection.
2. Moves the arm to the side-staging location (move to configuration with orientation path constraints to keep the fork straight).
3. Waits for a face to be detected.
4. Moves the face mesh in the planning scene to the detected mouth center position.
5. Computes the pose goal for the robot to move towards from the detected mouth center.
6. Moves the robot to that pose, with the orientation path constraint to keep the fork straight.
7. Toggles off face detection.

In order to implement this, this PR also makes the following changes to existing behaviors, decorators, and trees:
1. Change permissions on the feedback for the action from `EXCLUSIVE_WRITE` to `WRITE` to account for the fact that one tree may have multiple move_to actions.
    1. Note that the feedback may render poorly on the app to have multiple "thinking/moving" progress bars. We may need to come up with another way of rendering progress for actions with multiple robot motions.
2. Adds a `tree_root_name` parameter to every `MoveToTree`, because we were ended up with trees with multiple nested namespaces for behaviors, but the `MoveTo` behaviors must write to the top-level tree namespace in order for it to be read. Therefore, each tree's `name` parameter can be as nested as one wants, but the `tree_root_name` parameter must be the same.
3. Adds an optional `keys_to_not_write_to_blackboard` set to every MoveToTree, to allow us to specify certain keys that should not be hardcoded into the blackboard when initializing the tree, and will instead be written to the blackboard by another behavior.
4. Improves error handling:
    1. Allows `create_action_servers.py` to not crash when there are arbitrary exceptions in the trees.
    2. Allow `move_to` to deal with the scenario where a call to MoveIt finishes before a second tick of the tree.

# Testing procedure

Pull the latest code in [pymoveit2 branch `amaln/allowed_collision_matrix`](https://github.com/personalrobotics/pymoveit2/tree/amaln/allowed_collision_matrix).

As of now, this PR has only been tested in sim, with dummy face detection (waiting on #36 ). **Before merging, it must be tested on the real robot, with actual face detection**.

## Initialization
1. Ensure you are on the right branch of each of the dependencies and have the latest code. [PRL fork of pymoveit (branch: `amaln/allowed_collision_matrix`)](https://github.com/personalrobotics/pymoveit2), [PRL fork of py_trees_ros (branch: `amaln/service_client`)](https://github.com/personalrobotics/py_trees_ros/tree/amaln/service_client), and [feeding_web_interface (branch: `main`)](https://github.com/personalrobotics/feeding_web_interface).
2. Build your workspace: `colcon build`
3. Run the code.
    1. Sim
        1. Launch the F/T Sensor: `ros2 run ada_feeding dummy_ft_sensor.py`
        2. Launch the Dummy RealSense Node: `ros2 run feeding_web_app_ros2_test DummyRealSense`
        3. Launch Dummy FaceDetection: `ros2 run feeding_web_app_ros2_test FaceDetection`
        4. Launch the feeding nodes: `ros2 launch ada_feeding ada_feeding_launch.xml`
        5. Launch MoveIt: `ros2 launch ada_moveit demo_feeding.launch.py sim:=mock`
    2. Real Robot
        1. Launch the F/T Sensor: `ros2 run forque_sensor_hardware forque_sensor_hardware --ros-args -p host:=xxx.xxx.x.xx` (replace the x's with the correct IP)
        2. Launch the RealSense: See [the README in `ada_feeding_perception`](https://github.com/personalrobotics/ada_feeding/tree/ros2-devel/ada_feeding_perception)
        3. Launch Face Detection: `ros2 launch ada_feeding_perception ada_feeding_perception_launch.xml`
        4. Launch the feeding nodes: `ros2 launch ada_feeding ada_feeding_launch.xml`
        5. Launch MoveIt: `ros2 launch ada_moveit demo_feeding.launch.py`

## Testing
Reset the robot to the resting position: `ros2 action send_goal /MoveToRestingPosition ada_feeding_msgs/action/MoveTo "{}" --feedback` (you must do this every time before calling `MoveToMouth`). Then call the `MoveToMouth` action: `ros2 action send_goal /MoveToMouth ada_feeding_msgs/action/MoveTo "{}" --feedback`. Verify the following, in both sim and real.

- [x] It successfully runs the completion. This includes that it:
    - [x] Toggles on face detection.
    - [x] Moves the arm to the side-staging location.
    - [x] Waits for a face to be detected.
    - [x] Moves the face mesh in the planning scene to the detected mouth center position.
    - [x] Moves the robot to in front of the mouth.
    - [x] Toggles off face detection.
    - [x] Throughout all motions, the fork is kept mostly straight.
- [ ] You can terminate it at any of the above stages and it terminates cleanly.
    - [ ] Toggles on face detection.
    - [ ] Moves the arm to the side-staging location.
    - [ ] Waits for a face to be detected.
    - [ ] Moves the face mesh in the planning scene to the detected mouth center position.
    - [ ] Moves the robot to in front of the mouth.
    - [ ] Toggles off face detection.
    - [ ] Throughout all motions, the fork is kept mostly straight.

# Future Work
Currently, on the final approach to the user's mouth, the robot moves to a position 5cm away from their detected mouth center in the y direction of the base link (perpendicular to the wheelchair back), and an orientation within 0.5 radians of pointing straight at the wheelchair back. However:
1. Ideally, the fork should be perpendicular to the user's face, not to the wheelchair back. So if their face/mouth is rotated, the fork should also rotate accordingly (this requires FaceDetection perceiving the orientation of the user's face).
2. The fork should not be restricted to moving to a single point 5 cm in front of the user's mouth. Rather, it should be able to move anywhere that distance away from the mouth, within a range of angles (think in terms of spherical coordiantes centered on the mouth). In order to do this, we need to add many options for goal constraints to the MoveIt planning call. 

# Before Merging
- [ ] `Squash & Merge`

### DIFF
--- a/ada_feeding/config/ada_planning_scene.yaml
+++ b/ada_feeding/config/ada_planning_scene.yaml
@@ -3,7 +3,7 @@ ada_planning_scene:
   ros__parameters:
     object_ids: # list of objects to add to the planning scene
       - wheelchair
-      - wheelchair_collision
+      # - wheelchair_collision
       - table
       - head
     # For each object, specify the filename of the mesh, and the position and
--- a/ada_feeding/config/ada_planning_scene.yaml
+++ b/ada_feeding/config/ada_planning_scene.yaml
@@ -3,7 +3,7 @@ ada_planning_scene:
   ros__parameters:
     object_ids: # list of objects to add to the planning scene
       - wheelchair
-      # - wheelchair_collision
+      - wheelchair_collision
       - table
       - head
     # For each object, specify the filename of the mesh, and the position and
--- a/.pylintrc
+++ b/.pylintrc
@@ -25,13 +25,15 @@ clear-cache-post-run=no
 # A comma-separated list of package or module names from where C extensions may
 # be loaded. Extensions are loading into the active Python interpreter and may
 # run arbitrary code.
-extension-pkg-allow-list=cv2
+extension-pkg-allow-list=cv2,
+                         tf2_py
 
 # A comma-separated list of package or module names from where C extensions may
 # be loaded. Extensions are loading into the active Python interpreter and may
 # run arbitrary code. (This is an alternative name to extension-pkg-allow-list
 # for backward compatibility.)
-extension-pkg-whitelist=cv2
+extension-pkg-whitelist=cv2,
+                        tf2_py
 
 # Return non-zero exit code if any of these messages/categories are detected,
 # even if score is above --fail-under value. Syntax same as enable. Messages
--- a/ada_feeding/CMakeLists.txt
+++ b/ada_feeding/CMakeLists.txt
@@ -12,7 +12,6 @@ find_package(rcl_interfaces REQUIRED)
 find_package(moveit_msgs REQUIRED)
 find_package(pymoveit2 REQUIRED)
 find_package(std_srvs REQUIRED)
-find_package(py_trees_ros REQUIRED)
 # uncomment the following section in order to fill in
 # further dependencies manually.
 # find_package(<dependency> REQUIRED)
--- a/ada_feeding/CMakeLists.txt
+++ b/ada_feeding/CMakeLists.txt
@@ -11,6 +11,8 @@ find_package(ament_cmake_python REQUIRED)
 find_package(rcl_interfaces REQUIRED)
 find_package(moveit_msgs REQUIRED)
 find_package(pymoveit2 REQUIRED)
+find_package(std_srvs REQUIRED)
+find_package(py_trees_ros REQUIRED)
 # uncomment the following section in order to fill in
 # further dependencies manually.
 # find_package(<dependency> REQUIRED)
--- a/ada_feeding/README.md
+++ b/ada_feeding/README.md
@@ -19,8 +19,8 @@ This code has been developed and tested with the Kinova JACO Gen2 Arm, on comput
     1. Dummy node: `ros2 run ada_feeding dummy_ft_sensor.py`
     2. Real node: Follow the instructions in the [`forque_sensor_hardware` README](https://github.com/personalrobotics/forque_sensor_hardware/blob/main/README.md). Note that this is in the _main_ branch, which may not be the default branch.
 4. Launch MoveIt2:
-    1. Sim (RVIZ): `ros2 launch ada_moveit demo.launch.py sim:=mock`
-    2. Real Robot: `ros2 launch ada_moveit demo.launch.py`
+    1. Sim (RVIZ): `ros2 launch ada_moveit demo_feeding.launch.py sim:=mock`
+    2. Real Robot: `ros2 launch ada_moveit demo_feeding.launch.py`
 5. Launch the RealSense & Perception:
     1. Dummy nodes: `ros2 launch feeding_web_app_ros2_test feeding_web_app_dummy_nodes_launch.xml run_motion:=false run_web_bridge:=false`
     2. Real nodes: Follow the instructions in the [`ada_feeding_perception` README](https://github.com/personalrobotics/ada_feeding/blob/ros2-devel/ada_feeding_perception/README.md#usage)
--- a/ada_feeding/README.md
+++ b/ada_feeding/README.md
@@ -18,11 +18,14 @@ This code has been developed and tested with the Kinova JACO Gen2 Arm, on comput
 3. Launch the force-torque sensor:
     1. Dummy node: `ros2 run ada_feeding dummy_ft_sensor.py`
     2. Real node: Follow the instructions in the [`forque_sensor_hardware` README](https://github.com/personalrobotics/forque_sensor_hardware/blob/main/README.md). Note that this is in the _main_ branch, which may not be the default branch.
-4. Run the action servers: `ros2 launch ada_feeding ada_feeding_launch.xml`
-5. Launch MoveIt2:
-    1. RVIZ: `ros2 launch ada_moveit demo_feeding.launch.py sim:=mock`
-    2. Real Robot: `ros2 launch ada_moveit demo_feeding.launch.py`
-6. Test it:
+4. Launch MoveIt2:
+    1. Sim (RVIZ): `ros2 launch ada_moveit demo.launch.py sim:=mock`
+    2. Real Robot: `ros2 launch ada_moveit demo.launch.py`
+5. Launch the RealSense & Perception:
+    1. Dummy nodes: `ros2 launch feeding_web_app_ros2_test feeding_web_app_dummy_nodes_launch.xml run_motion:=false run_web_bridge:=false`
+    2. Real nodes: Follow the instructions in the [`ada_feeding_perception` README](https://github.com/personalrobotics/ada_feeding/blob/ros2-devel/ada_feeding_perception/README.md#usage)
+6. Run the action servers: `ros2 launch ada_feeding ada_feeding_launch.xml`
+7. Test it:
     1. Test the individual actions with the command line interface:
         1. `ros2 action send_goal /MoveAbovePlate ada_feeding_msgs/action/MoveTo "{}" --feedback`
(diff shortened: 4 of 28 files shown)


ada_feeding/config/ada_planning_scene.yaml
```diff
@@ -3,7 +3,7 @@ ada_planning_scene:
   ros__parameters:
     object_ids: # list of objects to add to the planning scene
       - wheelchair
-      - wheelchair_collision
+      # - wheelchair_collision
```
amalnanavati, line 6: TODO: Remove this once we've addressed [the issue with `wheelchair_collision` messing up certain planning calls](https://github.com/personalrobotics/ada_feeding/pull/39#issuecomment-1656435835).

amalnanavati:
Note: If you follow the instructions in the README, you need the [latest code on `feeding_web_interface`](https://github.com/personalrobotics/feeding_web_interface/commit/40ae747bb289710d22ba8e0726ec0d8ce858a9ba)

ada_feeding/config/ada_planning_scene.yaml
amalnanavati: Resolved [here](https://github.com/personalrobotics/ada_feeding/pull/39#issuecomment-1661287243)

amalnanavati:
Note: It is possible for the app to get into the situation where the robot is at the above-plate position and then the app calls `MoveToMouth` (e.g., if the user decides to skip acquisition, e.g., if the fork already has food on it). In this case, the start position does not satisfy orientation constraints, and the plan will either fail or be a weird two-part plan (because `ompl_planning.yaml` [has MoveIt attempt to fix when the start state is out-of-constraints](https://github.com/personalrobotics/ada_ros2/blob/73c837537afd604d9108f2e31b5bdbb687136e32/ada_moveit/config/ompl_planning.yaml#L9).)

There are a few possible ways to address this:
1. Have the MoveToMouth action first check if the starting position satisfies the orientation constraints, and only then add the orientation constraints. (This requires adding another behavior to subscribe to `tf` and check whether orientation constraints are satisfied.)
2. Have the MoveToMouth action first try moving with the orientation constraints, and if that fails (for any reason) have it retry without orientation constraints.
3. Change the [`MoveTo` action interface](https://github.com/personalrobotics/ada_feeding/blob/ros2-devel/ada_feeding_msgs/action/MoveTo.action) to take in the app state it is being called from. Since the app states (theoretically) directly correspond to a robot arm position, that should give MoveToMouth all the info it needs to decide whether or not to add orientation constraints.

I'm leaning towards 3, but lmk what you think after reviewing it.

egordon:
@amalnanavati Note: the branch in the doc doesn't exist for pymoveit2, I'm assuming your latest branch `amaln/allowed_collision_matrix` instead.

amalnanavati:
Yes, you are right. The latest branch is documented [here](https://docs.google.com/drawings/d/1yRElxttZ7kord7RLDpK-xaIPzuqAWXYU4on7UYu9B64/edit). I also edited the PR comment

amalnanavati:
Rebased onto the new `ros2-devel`. Will test in-person, with face detection, tomorrow.

amalnanavati:
I re-ran all the sim tests and verified that it works (with the most up-to-date code in `ros2-devel` and `amaln/pre_moveto_idiom`). I will test it in-person on Mon, but if folks want to get a head-start on the review they can.

Raidakarim:
I pulled all new code and checked out the branches mentioned in the testing procedure.  I faced some obstacles as shown below when I tried testing in sim:
Seems like this command `Launch MoveIt: ros2 launch ada_moveit demo_feeding.launch.py sim:=mock` does not work. When I tried modifying it a bit like this `ros2 launch ada_moveit demo.launch.py sim:=mock`; it worked.
Screenshot 2023-08-29 at 1 46 06 PM
After launching all 5 terminals as listed in sim (a-e), I opened another terminal and ran `ros2 action send_goal /MoveToRestingPosition ada_feeding_msgs/action/MoveTo "{}" --feedback` got an error the launch file terminal and the move action terminal didn't run anything:
Screenshot 2023-08-29 at 12 26 20 PM
Screenshot 2023-08-29 at 12 27 51 PM
Then I modified the ceate_action_servers.py file's ada_watchdog_listener and ADAWatchdogListener to respectively be ada_watchdog and ADAWatchdog.  
Screenshot 2023-08-29 at 1 37 53 PM
Then, when I ran the move action for resting position command again, I got this error:
Screenshot 2023-08-29 at 12 46 59 PM
I didn't make any further changes.

amalnanavati:
@Raidakarim you need to have also pulled the most recent `ada_ros2` code. Also ensure you have the latest `pymoveit2` code on the `amaln/allow_collision_matrix` branch

Raidakarim:
> @Raidakarim you need to have also pulled the most recent `ada_ros2` code. Also ensure you have the latest `pymoveit2` code on the `amaln/allow_collision_matrix` branch

I followed Amal's suggestion. I really like the new arm with table and user. 
Screenshot 2023-08-29 at 6 00 03 PM

Then, launched all terminals in sim again and ran `ros2 action send_goal /MoveToRestingPosition ada_feeding_msgs/action/MoveTo "{}" --feedback`. But got a new timeout error:
Screenshot 2023-08-29 at 7 05 10 PM

amalnanavati:
@Raidakarim you have the old controller. You need to delete the `ros2_controllers` folder in your workspace, and instead clone the [`pr_ros_controllers` folder](https://github.com/personalrobotics/pr_ros_controllers/tree/main), and then rebuild your workspace from scratch.

Raidakarim:
> @Raidakarim you have the old controller. You need to delete the `ros2_controllers` folder in your workspace, and instead clone the [`pr_ros_controllers` folder](https://github.com/personalrobotics/pr_ros_controllers/tree/main), and then rebuild your workspace from scratch.

I looked into my workspace folders, and searched in the src, install folders, but could not find a ros2_controllers folder. Where can I find it?

amalnanavati:
@Raidakarim Please take this discussion onto Slack, off of this PR. Comments on this PR should focus on code feedback, not debugging separate issues.

amalnanavati:
TODO: For the real F/T sensor, we should look into whether the re-tare service returns during the 0.75 secs when the F/T readings are paused, or after. Because the current `pre_moveto_config` immediately toggles the watchdog listener after the re-taring service returns, but maybe there should be a sleep in-between.

taylorkf:
When I kill MoveToMouth during the arm movement to the staging position, reset to the resting position, and then attempt MoveToMouth again, I get this py_trees error:

[create_action_servers.py-5]     for node in self.root.tick():
[create_action_servers.py-5]   File "/opt/ros/humble/lib/python3.10/site-packages/py_trees/composites.py", line 425, in tick
[create_action_servers.py-5]     assert self.current_child is not None  # should never be true, help mypy out
[create_action_servers.py-5] AssertionError

taylorkf (approved):
Looks good. We still have an (infrequent) error where the robot does not approach a mouth even when it is detected and within 1.5 meters. We'll keep an eye out for this to happen again and track the error messages.

Raidakarim:
For Initialization Step 3i, in Sim point d, I had to set use_estop to false to the given command. So, the command I used to successfully ran it is: `ros2 launch ada_feeding ada_feeding_launch.xml use_estop:=false`

If I don't add the use_estop part and just use the current given command, it doesn't run successfully and gives errors. Also, I ran it in `ros2-devel` branch after this PR has been merged. So, this might not only be true for this PR but also for the `ros2-devel` branch.

amalnanavati:
Updated the README to account for that in [this commit](https://github.com/personalrobotics/ada_feeding/commit/bdfc4e3da3a9ba57e418795f1172b9f681a06b92).

PR #46, opened 2023-08-16
ROS launch topic remapping and action_type param

egordon:
# Description

This PR does some linting / cleanup, replaces the topic-in-rosparam antipattern with topic remapping in the ROS launch file, and passes action_type to `create_tree` as every single tree pretty much requires it. It removes a redundant param in the YAML.

# Testing procedure

[TODO: test on my end, but this should not have changed any functionality, so all previous tests are valid]

# Before Merging
- [ ] `Squash & Merge`

### DIFF
--- a/ada_feeding/ada_feeding/trees/move_to_configuration_tree.py
+++ b/ada_feeding/ada_feeding/trees/move_to_configuration_tree.py
@@ -27,7 +27,6 @@ class MoveToConfigurationTree(MoveToTree):
 
     def __init__(
         self,
-        action_type_class_str: str,
         joint_positions: List[float],
         tolerance: float = 0.001,
         weight: float = 1.0,
@@ -39,10 +38,6 @@ class MoveToConfigurationTree(MoveToTree):
 
         Parameters
         ----------
-        action_type_class_str: The type of action that this tree is implementing,
-            e.g., "ada_feeding_msgs.action.MoveTo". The input of this action
-            type can be anything, but the Feedback and Result must at a minimum
-            include the fields of ada_feeding_msgs.action.MoveTo
         joint_positions: The joint positions to move the robot arm to.
         tolerance: The tolerance for the joint positions.
         weight: The weight for the joint goal constraint.
--- a/ada_feeding/ada_feeding/trees/move_to_dummy_tree.py
+++ b/ada_feeding/ada_feeding/trees/move_to_dummy_tree.py
@@ -64,7 +64,7 @@ class MoveToDummyTree(ActionServerBT):
         Parameters
         ----------
         name: The name of the behavior tree.
-        action_type: full name of a Python class for the associated action, 
+        action_type: full name of a Python class for the associated action,
             can be converted to a type object with `import_from_string` in
             helpers.py
         logger: The logger to use for the behavior tree.
--- a/ada_feeding/ada_feeding/trees/move_to_dummy_tree.py
+++ b/ada_feeding/ada_feeding/trees/move_to_dummy_tree.py
@@ -37,8 +37,7 @@ class MoveToDummyTree(ActionServerBT):
         dummy_plan_time: How many seconds this dummy node should spend in planning.
         dummy_motion_time: How many seconds this dummy node should spend in motion.
         """
-        self.action_type_class = None
-
+        
         # Set the dummy motion parameters
         self.dummy_plan_time = dummy_plan_time
         self.dummy_motion_time = dummy_motion_time
--- a/ada_feeding/ada_feeding/trees/move_to_dummy_tree.py
+++ b/ada_feeding/ada_feeding/trees/move_to_dummy_tree.py
@@ -26,7 +26,6 @@ class MoveToDummyTree(ActionServerBT):
 
     def __init__(
         self,
-        action_type_class_str: str,
         dummy_plan_time: float = 2.5,
         dummy_motion_time: float = 7.5,
     ) -> None:
@@ -35,15 +34,10 @@ class MoveToDummyTree(ActionServerBT):
 
         Parameters
         ----------
-        action_type_class_str: The type of action that this tree is implementing,
-            e.g., "ada_feeding_msgs.action.MoveTo". The input of this action
-            type can be anything, but the Feedback and Result must at a minimum
-            include the fields of ada_feeding_msgs.action.MoveTo
         dummy_plan_time: How many seconds this dummy node should spend in planning.
         dummy_motion_time: How many seconds this dummy node should spend in motion.
         """
-        # Import the action type
-        self.action_type_class = import_from_string(action_type_class_str)
+        self.action_type_class = None
 
         # Set the dummy motion parameters
         self.dummy_plan_time = dummy_plan_time
@@ -56,6 +50,7 @@ class MoveToDummyTree(ActionServerBT):
     def create_tree(
         self,
         name: str,
+        action_type: str,
         logger: logging.Logger,
         node: Node,
     ) -> py_trees.trees.BehaviourTree:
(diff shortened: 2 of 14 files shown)


ada_feeding/ada_feeding/trees/move_to_dummy_tree.py
```diff
@@ -35,15 +34,10 @@ def __init__(
...
-        action_type_class_str: The type of action that this tree is implementing,
-            e.g., "ada_feeding_msgs.action.MoveTo". The input of this action
-            type can be anything, but the Feedback and Result must at a minimum
-            include the fields of ada_feeding_msgs.action.MoveTo
         dummy_plan_time: How many seconds this dummy node should spend in planning.
         dummy_motion_time: How many seconds this dummy node should spend in motion.
         """
-        # Import the action type
-        self.action_type_class = import_from_string(action_type_class_str)
+        self.action_type_class = None
```
amalnanavati, line 40: Is this line necessary? It's a departure from the logic used in other MoveTo trees.

ada_feeding/ada_feeding/trees/move_to_tree.py
```diff
@@ -31,24 +31,16 @@ class MoveToTree(ActionServerBT, ABC):
 
     def __init__(
         self,
-        action_type_class_str: str,
     ) -> None:
-        """
-        Initializes tree-specific parameters.
```
amalnanavati, line 37: I'm not a fan of having no docstring on a function. If we ever automatically generate documentation, then a function without a docstring would have nothing.

Can you keep the docstring and just remove the decription of `action_type_class_str`?

ada_feeding/ada_feeding/trees/move_to_dummy_tree.py
amalnanavati: Note (if you did it for the linter): idk I'm okay ignoring the linter warning about initializing an attribute outside of the the __init__ function. Where it makes sense I think we should, but there are other cases where it seems unnecessary, like here.

ada_feeding/ada_feeding/trees/move_to_tree.py
```diff
@@ -31,24 +31,16 @@ class MoveToTree(ActionServerBT, ABC):
...
-        action_type_class_str: The type of action that this tree is implementing,
-            e.g., "ada_feeding_msgs.action.MoveTo". The input of this action
-            type can be anything, but the Feedback and Result must at a minimum
-            include the fields of ada_feeding_msgs.action.MoveTo
-        """
-        # Import the action type
-        self.action_type_class = import_from_string(action_type_class_str)
+        # Uninitialized attributes
+        self.action_type_class = None
+        self.action_type_class_str = ""
```
amalnanavati, line 37: Looks like these two initializations are not necessary -- I checked all files that subclass this, and they only access `self.action_type_class_str` in `create_move_to_tree`, which is called after these parameters are initialized in `create_tree`.

Was this for the linter? See my above comment re. that. My marginal preference is to not unnecessarily initialize variables in `__init__`, but I don't feel super strongly about it so am also okay with this option if you prefer. We should just be on the same page to have a uniform coding style in this repo.

ada_feeding/launch/ada_feeding_launch.xml
```diff
@@ -11,11 +11,13 @@
   <!-- Launch the watchdog -->
   <node pkg="ada_feeding" exec="ada_watchdog.py" name="ada_watchdog">
     <param from="$(find-pkg-share ada_feeding)/config/ada_watchdog.yaml"/>
+    <remap from="~/ft_topic" to="/wireless_ft/ftSensor1" />
   </node>
 
   <!-- Launch the action servers necessary to move the robot -->
   <node pkg="ada_feeding" exec="create_action_servers.py" name="ada_feeding_action_servers">
     <param from="$(find-pkg-share ada_feeding)/config/ada_feeding_action_servers.yaml"/>
+    <remap from="~/watchdog" to="/ada_watchdog/watchdog" />
```
amalnanavati, line 20: Ah, using relative topics (like you did here) is better than absolute topics ([like I did in SegmentFromPoint](https://github.com/personalrobotics/ada_feeding/blob/b251e15ecac9edd49f595d84d920b3d087a75b48/ada_feeding_perception/launch/ada_feeding_perception_launch.xml#L7)). If you want to fix it in this PR (in both the launchfile and the ROS node) you're welcome to, otherwise I'll do it.

ada_feeding/scripts/ada_planning_scene.py
```diff
@@ -64,7 +64,10 @@ def load_parameters(self) -> None:
             descriptor=ParameterDescriptor(
                 name="assets_dir",
                 type=ParameterType.PARAMETER_STRING,
-                description="The absolute path to the directory to find the assets (e.g., STL mesh files).",
+                description=(
+                    "The absolute path to the directory to find the assets "
+                    "(e.g., STL mesh files)."
```
amalnanavati, line 69: JW for formatting did you use Black Formatter (`python3 -m black .`) or something else? And if black, did you add any flags to it?

We should just ensure that we're using the same formatter to avoid successive PRs making/reverting style changes. I use `python3 -m black .`

ada_feeding/scripts/create_action_servers.py
```diff
@@ -287,10 +302,10 @@ def create_action_servers(
             try:
                 assert issubclass(
                     self._tree_classes[params.tree_class], ActionServerBT
-                ), ("Tree %s must subclass ActionServerBT" % params.tree_class)
-            except Exception:
+                ), f"Tree {params.tree_class} must subclass ActionServerBT"
+            except AssertionError:
```
amalnanavati, line 306: Good catch

amalnanavati (approved):
LGTM. All comments are small, feel free to incorporate or ignore them.

ada_feeding/launch/ada_feeding_launch.xml
amalnanavati: Coming to think of it, I think there needs to be a separate version of this PR for `ada_feeding_perception` (including to address linter warnings), so I can be in charge of that.

ada_feeding/ada_feeding/trees/move_to_dummy_tree.py
egordon: I thought I did it for the linter, but it's not complaining. Removed.

ada_feeding/ada_feeding/trees/move_to_tree.py
egordon: Done

ada_feeding/ada_feeding/trees/move_to_tree.py
egordon: I thought it was for the linter, but it's not complaining, removed.

ada_feeding/scripts/ada_planning_scene.py
egordon: No flags w/ black. This was a compromise between black and pylint that was complaining about the long lines.

ada_feeding/ada_feeding/trees/move_to_tree.py
```diff
@@ -67,7 +62,12 @@ def create_tree(
         tree: The behavior tree that moves the robot above the plate.
         """
         self.create_blackboard(name)
-        return self.create_move_to_tree(name, logger, node)
+
+        # Import the action type
+        self.action_type_class = import_from_string(action_type_class_str)
```
amalnanavati, line 67: Not sure how I didn't catch this, but I think this breaks the code (it should be `action_type`, not `action_type_class_str`). I'm currently working on a large linting PR and will fix this as part of that.

ada_feeding/ada_feeding/trees/move_to_configuration_tree.py
amalnanavati, line 1: On line 49 also I think `action_type_class_str` should have gotten removed. I'm surprised this code passed tests. Anyway, I'll change it on my end.

ada_feeding/ada_feeding/trees/move_to_tree.py
amalnanavati: Addressed in #47 , I will test it and then tag you for review.

ada_feeding/ada_feeding/trees/move_to_configuration_tree.py
amalnanavati: Addressed in #47 , I will test it and then tag you for review.

PR #48, opened 2023-08-18
Add Node that kills itself when the watchdog fails

amalnanavati:
# Description

This PR does the following:

1. Separates the watchdog listener logic from `create_action_servers.py` to its own class.
2. Creates a node that kills itself if the watchdog fails. This node exposes a service that allows it to be toggled on and off.
    1. As part of that, I changed the watchdog listener behavior to set the default `watchdog_failed` to False. This means that if the watchdog is not publishing when the listener starts listening, it will take 0.5sec (configureable by a param) before the watchdog listener registers a failure (because that is how long we allow the watchdog to go without publishing). The reason for this change is so that if we create a watchdog listener and query it before a watchdog message has been received (e.g., if it is listening at a faster rate than the watchdog), it doesn't immediately think the watchdog failed.
3. Modifies `ada_planning_scene.py` to wait until MoveIt is running before publishing the messages to populate the planning scene. This is because of an ordering dependency: `ada_feeding_launch.xml` launches both the watchdog and `ada_planning_scene.py`. But the MoveIt launch requires the watchdog to be running before it is launched, and `ada_planning_scene.py` requires MoveIt to be running before it is launched. This change resolves the ordering issue.
4. Modifies the dummy F/T sensor to expose a re-taring service, that stops it from publishing messages for 0.75 sec (to mimic the fact that the actual F/T sensor stops publishing messages for that long when re-taring).

# Testing procedure

Pull the code from [ada_ros2#10](https://github.com/personalrobotics/ada_ros2/pull/10).

1. Verify that `create_action_servers.py` still works as expected:
    - [x] Run the dummy FT sensor (`ros2 run dummy_ft_sensor.py`), the ada_feeding launchfile, and MoveIt in RVIZ. Send a MoveAbovePlate action and verify that it succeeds.
    - [x] Kill the dummy FT sensor. Send a MoveAbovePlate action and verify that it is rejected.
    - [x] Re-start the dummy FT sensor. Send a MoveAbovePlate action, and kill the dummy FT sensor before the action completes. Verify that the action aborts.
3. Verify that re-taring the F/T sensor works properly. 
    - [x] Run the dummy F/T sensor, echo the F/T topic, and re-tare the sensor (`ros2 service call /wireless_ft/set_bias std_srvs/srv/SetBool "{data: true}"`). Verify that the messages stop publishing for 0.75 sec.

# Before Merging
- [ ] `Squash & Merge`


amalnanavati:
Soo... re-taring causes the FT sensor to stop publishing for ~0.75 seconds. During that time, this watchdog listener trips and kills controllers.

Taylor and I brainstormed a couple of options and the one that seemed the most promising is to add a service call that toggles on/off this node (the one that kills itself upon watchdog failure). Only while the node is on will it kill itself if the watchdog fails. And then, in every MoveTo behavior tree, we first retare the FT sensor, then toggle this node on, and then move (I can build an idiom for this). Thoughts @egordon ?

To be clear, this would not involve changing the watchdog itself, only the watchdog listener that kills controllers when the watchdog fails.

amalnanavati:
Ready for review. I went ahead and implemented the service to toggle the listener on and off (explained above), and modified the description and testing procedure accordingly.

ada_feeding/ada_feeding/ada_watchdog_listener.py
```diff
@@ -0,0 +1,97 @@
...
+
+# pylint: disable=too-few-public-methods
+# This class only needs one public method to check if the watchdog is ok.
+class ADAWatchdogListener:
+    """
+    The ADAWatchdogListener class listens to the watchdog topic and can be queried
+    to determine if the watchdog is still alive.
+    """
+
+    def __init__(self, node: Node):
```
egordon: Not necessary, but consider adding an event-based API to this as well. i.e. init with an optional callback function (that can be changed later). And once the watchdog fails, call that function with the Diagnostic.

In a separate function to set/change the callback function, don't allow it to happen if the watchdog has already failed.

ada_feeding/scripts/dummy_ft_sensor.py
```diff
@@ -89,12 +105,34 @@ def __init__(self, rate_hz: float = 30.0) -> None:
...
+        return response
+
     def publish_msg(self) -> None:
         """
         Publish a message to the force-torque sensor topic.
         """
         # Only publish if the sensor is on
         if self.get_parameter("is_on").value:
+            # Don't publish for 0.75 sec after a set_bias request
+            with self.set_bias_request_time_lock:
```
egordon, line 133: Points for realism!

ada_feeding/scripts/ada_watchdog_listener_node.py
```diff
@@ -0,0 +1,107 @@
...
+                type=ParameterType.PARAMETER_DOUBLE,
+                description=(
+                    "The rate (Hz) at which to check the whether the watchdog has failed."
+                ),
+                read_only=True,
+            ),
+        )
+
+        # Create a service to toggle this node on-and-off
+        self.toggle_service = self.create_service(
```
egordon, line 37: What's the purpose of this service? I don't think it's bad to have it necessarily, but I do have an issue with the default being that this node is not on and requiring a specific activation.

By default, this node should be on if it is running, and require the user to be _explicit_ when turning it off (either, e.g., setting `watchdog:=false` in the launch file or explicitly calling a service to disable it).

ada_feeding/config/ada_feeding_action_servers.yaml
```diff
@@ -2,7 +2,7 @@
 ada_feeding_action_servers:
   ros__parameters:
     # If we haven't received a message from the watchdog in this time, stop all robot motions
-    watchdog_timeout_sec: 0.5
+    watchdog_timeout: 0.5
```
egordon: If not in the variagble name, add `# seconds` or `# s` afterwards to specify unit.

egordon (approved):
This is fine for now (with comments addressed), but looking at this, I think the ADAWatchdogListener and Node might actually make more sense in `ada_ros2`, so we don't have any of those weird circular dependencies (i.e. `ada_ros2` ideally shouldn't have any reference to `ada_feeding`, and there is nothing feeding-specific about specifically the watchdog listener, even if the watchdog generator is indeed feeding-specific).

Consider making an issue.

ada_feeding/scripts/ada_watchdog_listener_node.py
taylorkf: So the idea is that since re-taring causes the sensor to stop publishing for 0.75 sec, we could only have the watchdog running when movement is about to occur after re-taring (described in Amal's comment above). However, it's probably a better idea (if possible) to have the ability to toggle the watchdog off and back on around re-taring, rather than having it off as a default. @amalnanavati thoughts?

amalnanavati: I agree that it should be on by default. Before any MoveTo Behavior, we should do the following:

1. Turn the watchdog listener off.
2. Re-tare the F/T sensor.
3. (Optional) Set F/T thresholds for that motion.
4. Turn the watchdog listener on.

I'll define an idiom that does that.

ada_feeding/config/ada_feeding_action_servers.yaml
amalnanavati: Agreed. I've increasingly been adding the units to the end of the parameter name because if people define a custom config file, then the comment might go away. But I haven't yet uniformly enforced that covention.

amalnanavati:
The above commits address PR comments. I re-ran all tests and verified that it works.

Next, I'll move the watchdog listener functionality into `ada_ros2`

amalnanavati:
All tested and works. Will merge.

PR #50, opened 2023-08-24
Implement the E-Stop Condition for the Watchdog

amalnanavati:
# Description

This PR implements the e-stop button for the watchdog, and streamlines the code for watchdog conditions:

1. Created an abstract class, `WatchdogCondition`, that all watchdog conditions must implement. Every watchdog condition must have one or more startup conditions and one or more status conditions. Whenever either the startup or status is checked, it returns a list containing whether the condition passed or failed and a description of the condition.
2. Moved the `FTSensorCondition` to inherit from the `WatchdogCondition` class.
    1.  **Startup Condition(s)**: Has received at least one message on the FT topic.
    2. **Status Condition(s)**: Every dimension of the wrench message has received a unique value in the last 0.5 secs (configureable by parameter). In practice, this requires at least one message every 0.5 secs which have none of the same values as the previous message.
3. Implemented the `EStopCondition` Class.
    1.  **Startup Condition(s)**: Has been clicked at least once (to ensure it is plugged in).
    2. **Status Condition(s)**: (a) Has not been unplugged (which the node checks by reading from the ACPI event socket). (b) Has not been clicked since startup (in practice, this requires the button to be clicked exactly once -- for startup -- and not again).
        1. This node detects a "button click" if there is a rising edge that passes `max_threshold` or a falling edge that passes `min_threshold`. All such detections within a 0.5 sec period are considered part of the same button click.
4. Modified `ADAWatchdog` to use the above two conditions.

# Testing procedure

Tested the following on `t0b1`. Note that when echoed, the success byte appears as `\0` and the error byte occurs as `\x02`

- [x] 1. Plug in the e-stop button, and launch the watchdog (`ros2 launch ada_feeding ada_feeding_launch.xml`). Echo the watchdog messages. Ensure that the conditions that show up are the startup conditions, and none of them pass.
- [x] 2. Click the e-stop button. Ensure that the e-stop startup condition passes, but the FT startup condition still fails.
- [x] 3. [Run the real F/T sensor](https://github.com/personalrobotics/forque_sensor_hardware/blob/main/README.md#sample-usage). Ensure that the FT startup condition passes. The echoed watchdog messages should now move to the regular conditions. Ensure that all are passing (in other words, the e-stop button click was not mis-represented as two button clicks).
- [x] 4. Press the e-stop button. Ensure the corresponding e-stop condition fails.
- [x] 5. Terminate the F/T node. Ensure the corresponding condition fails.
- [x] 6. Re-run 1-3. Unplug the e-stop button. Ensure the corresponding condition fails.
- [x] 7. Repeat step 1 and run the FT sensor. Repeat step 2 with a variety of edge-case button presses (e.g., soft press, hard press, etc.). Ensure that the startup condition passes, but the regular condition still succeeds (e.g., a single button press is not mis-detected as multiple).
- [x] 8. Repeat step 1 and run the FT sensor. Wait 2 secs since startup (because the first 2 sec of e-stop data is ignored due to initial noise). Hold down the e-stop button for > 0.5 sec. Ensure the startup condition passes. Let go of the e-stop button. Ensure the regular condition fails.
- [x] 9. Repeat steps 1-3. Try a variety of e-stop button presses (soft press, hard press, long press, press at non-central location of button, etc.). Ensure that the condition fails for every press.

# Before Merging
- [x] `Squash & Merge`


amalnanavati:
- Rebased onto `ros2-devel` now that the previous commits in this branch chain have been merged in.
- Tested on the real robot with the e-stop button and verified that it works.

Ready for review!

taylorkf:
Quick comment that the pyaudio version needs to be >= 0.2.12, which can be done by installing through pip.

ada_feeding/ada_feeding/watchdog/ft_sensor_condition.py
```diff
@@ -0,0 +1,180 @@
...
+
+        Returns
+        -------
+        startup_status: A list of tuples, where each tuple contains a boolean
+            status of a startup condition and a string describing the condition.
+            All conditions must be True for the startup condition to be considered
+            passed. For example, [(False, "Has received at least one message on
+            topic X")] means that the startup condition has not passed because
+            the node has not received any messages on topic X yet.
+        """
```
taylorkf, line 141: We discussed changing these messages to "Has not received at least one message on topic X" and "E-stop button has not been clicked at least once" when the conditions aren't satisfied. This will make for easier readability on the app side.

ada_feeding/ada_feeding/watchdog/estop_condition.py
```diff
@@ -0,0 +1,421 @@
...
+                    "a button click"
+                ),
+                read_only=True,
+            ),
+        )
+        self.max_threshold = max_threshold.value
+
+        time_per_click_sec = self._node.declare_parameter(
+            "time_per_click_sec",
+            0.5,
```
taylorkf, line 198: Not something that needs to be changed now, but let's keep an eye out during testing to make sure that 0.5 sec is enough time for most/all users.

taylorkf (approved):
Looks good. Needs testing on Lovelace to make sure that the cable connections work, but the code is solid and it has been well tested.

PR #51, opened 2023-08-25
Hotfix: Actually remove watchdog_listener from `ada_feeding`

amalnanavati:
#48 was missing the final commit that actually removed watchdog_listeners from it. This PR addresses that. It has been tested with the same tests in #48.


amalnanavati:
Tested with the tests from #48 and verified that it works. Bypassing approval since this should have been merged with the previous PR.

PR #52, opened 2023-08-25
Add Pre MoveTo Idiom

amalnanavati:
# Description

Before any call to MoveTo, we should do the following: (a) toggle the `watchdog_listener` off; (b) re-tare the F/T sensor; (c) toggle the `watchdog_listener` on; and (d) set force-torque thresholds. The reason for (a) and (c) is that re-taring causes the force-torque sensor to momentarily stop publishing, which kills the watchdog listener.

In order to make it easy to add those calls before any MoveTo node, this PR does the following:

1. Creates an idiom, `pre_moveto_config` that does the above 4 steps.
2. In service of the above, creates an idiom, `retry_call_ros_service`, to make it easier to call ros services and verify the response.
3. Implements the `MoveToConfigurationWithFTThresholdsTree` that adds the `pre_moveto_config` before a `MoveToConfiguration` call.
4. Implements `MoveToRestingPosition`, `MoveAbovePlate`, and `MoveToStowLocation` as `MoveToConfigurationWithFTThresholdsTree`.

# Testing procedure

1. Pull the latest [py_trees_ros (branch: `amaln/service_client`) code](https://github.com/personalrobotics/py_trees_ros/tree/amaln/service_client) and re-run `colcon build`
2. Start the F/T sensor: `ros2 run forque_sensor_hardware forque_sensor_hardware --ros-args -p host:=xxx.xxx.x.xx`
3. Echo the F/T sensor topic: `ros2 topic echo /wireless_ft/ftSensor1`
4. Start the ADA feeding nodes: `ros2 launch ada_feeding ada_feeding_launch.xml`
5. Run MoveIt with the Watchdog Listener: `ros2 launch ada_moveit demo_feeding.launch.py`
6. Verify the following:
    - [x] The F/T sensor readings are not mean 0 (this is common if it hasn't yet been re-tared).
    - [x] Run `ros2 action send_goal /MoveAbovePlate ada_feeding_msgs/action/MoveTo "{}" --feedback`
        - [x] Verify that the MoveIt launchfile prints the messages that indicate that the watchdog listener was turned off, and then turned back on.
        - [x] Verify that the F/T sensor readings become mean-0 (a sign that it re-tared).
        - [x] Verify that the MoveAbovePlate action succeeds.

# Before Merging
- [x] `Squash & Merge`


ada_feeding/ada_feeding/trees/move_to_configuration_with_ft_thresholds_tree.py
```diff
@@ -0,0 +1,160 @@
...
+                tolerance=self.tolerance_joint,
+                weight=self.weight_joint,
+                planner_id=self.planner_id,
+                allowed_planning_time=self.allowed_planning_time,
+            )
+            .create_tree(name, self.action_type, logger, node)
+            .root
+        )
+
+        # Add the pose path constraints
```
amalnanavati:
```suggestion
        # Add the re-taring and FT thresholds
```

ada_feeding/ada_feeding/idioms/retry_call_ros_service.py
```diff
@@ -0,0 +1,133 @@
...
+
+def retry_call_ros_service(
+    name: str,
+    service_type: Any,
+    service_name: str,
+    key_request: Optional[str] = None,
+    request: Optional[Any] = None,
+    key_response: Optional[str] = None,
+    response_checks: Optional[List[py_trees.common.ComparisonExpression]] = None,
+    max_retries: int = 3,
```
taylorkf, line 27: Is there a reason for max_retries to be 3 here? Would we want to change it in the future?

ada_feeding/ada_feeding/idioms/retry_call_ros_service.py
```diff
@@ -0,0 +1,133 @@
...
+    wait_for_server_timeout_sec: float = 3.0,
```
taylorkf, line 28: Same here, should wait_for_server_timeout_sec be something that we have easy access to change? If not, you can ignore this comment.

ada_feeding/ada_feeding/idioms/retry_call_ros_service.py
taylorkf: If not, you can ignore this comment.

ada_feeding/ada_feeding/idioms/pre_moveto_config.py
```diff
@@ -0,0 +1,226 @@
...
+        2. Set force-torque thresholds (so the ForceGateController will trip if
+           the force-torque sensor readings exceed the thresholds).
+
+    Parameters
+    ----------
+    name: The name to associate with this behavior.
+    re_tare: Whether to re-tare the force-torque sensor.
+    toggle_watchdog_listener: Whether to toggle the watchdog listener on and off.
+        In practice, if the watchdog listener is on, you should toggle it.
+    f_mag: The magnitude of the overall force threshold. No threshold if 0.0.
```
taylorkf, line 75: A bit confused about these statements...is this saying that if f_mag=0.0 then it signifies no threshold?

taylorkf (approved):
Works as expected after removing OctoMap. Just a few comments on readability.

ada_feeding/ada_feeding/idioms/pre_moveto_config.py
amalnanavati: That is correct. The current default (all 0.0) basically keeps moving irrespective of the forces/torques. [This is how the ForceGateController is defined](https://github.com/personalrobotics/pr_ros_controllers/blob/f3c11af37a2c53e8de5f5ea39c723f1e2f031d27/force_gate_controller/src/force_gate_controller.cpp#L164).

I had them all default to 0 because [that is what the ForceGateController does](https://github.com/personalrobotics/pr_ros_controllers/blob/main/force_gate_controller/src/force_gate_controller_parameters.yaml), but I am realizing now that that is not a good idea for feeding. I'll change `f_mag` to default to 4, which [is the default when ada_moveit is launched](https://github.com/personalrobotics/ada_ros2/blob/e1229060469d3ae17c2ada28030cf9f78715a70d/ada_moveit/config/real_controllers.yaml#L36).

ada_feeding/ada_feeding/idioms/retry_call_ros_service.py
amalnanavati: I just chose 3 because it seemed reasonable to me.

Any behavior tree that uses this idiom can pass a custom parameter to this function for a different value for max_retries. For example, we can change the code of `pre_moveto_config.py` to change the number of retries for each of its service calls. 

A question is whether we would want to expose that as a parameter to the behavior tree itself, which can then be defined via the yaml file. The benefit of that is that we can change the num retries without changing the code. But I'm not convinced that is necessary, because the main reason to retry on a service call is if the service isn't available, and beyond some number of retries it won't make sense to keep retrying. So I think we can probably hardcode a reasonable value per service call, and therefore don't need to expose the parameter all the way up to the level of the yaml file.

ada_feeding/ada_feeding/idioms/retry_call_ros_service.py
amalnanavati: See above comment, I think changing it within the tree that calls this idiom is sufficient, I don't think we need to expose it all the way up to a parameter of the tree (so it can be changed via yaml file)

ada_feeding/ada_feeding/idioms/pre_moveto_config.py
```diff
@@ -0,0 +1,226 @@
...
+    Checks that all the parameters were successfully set.
+    """
+    return all(result.successful for result in blackboard_value.results)
+
+
+def pre_moveto_config(
+    name: str,
+    re_tare: bool = True,
+    toggle_watchdog_listener: bool = True,
+    f_mag: float = 0.0,
```
amalnanavati:
```suggestion
    f_mag: float = 4.0,
```

PR #56, opened 2023-08-30
Added MoveFromMouth Actions

amalnanavati:
# Description

In service of #53 . At the end of transfer, the robot is colliding with the wheelchair collision mesh. Therefore, the actions that take the robot from that position to above the plate or the resting position need to toggle that mesh off.

These actions for now execute the entire motion with the mesh toggled off. However, in the future we should consider only executing an initial part of the motion (e.g., move back to the staging location) with the mesh off, then re-enable it, and then complete the rest of the motion, for increased safety.

# Testing procedure

Pull the code from [ada_ros2#12](https://github.com/personalrobotics/ada_ros2/pull/12).

- [x] On the real robot, call MoveToMouth as documented in #42 . 
- [x] Run `ros2 action send_goal /MoveAbovePlate ada_feeding_msgs/action/MoveTo "{}" --feedback` and verify that it fails.
- [x] Run `ros2 action send_goal /MoveFromMouthToAbovePlate ada_feeding_msgs/action/MoveTo "{}" --feedback` and verify that it succeeds.
- [x] Run `ros2 action send_goal /MoveToRestingPosition ada_feeding_msgs/action/MoveTo "{}" --feedback` and verify that it fails.
- [x] Run `ros2 action send_goal /MoveFromMouthToRestingPosition ada_feeding_msgs/action/MoveTo "{}" --feedback` and verify that it succeeds.

# Before opening a pull request
- [x] Format your code using [black formatter](https://black.readthedocs.io/en/stable/) `python3 -m black .`
- [x] Run your code through [pylint](https://pylint.readthedocs.io/en/latest/) and address all warnings/errors. The only warnings that are acceptable to not address is TODOs that should be addressed in a future PR. From the top-level `ada_feeding` directory, run: `pylint --recursive=y --rcfile=.pylintrc .`.

# Before Merging
- [ ] `Squash & Merge`

Issue #65, opened 2023-08-31
[ROS2] Improve Collision Scene


amalnanavati:
Right now, our collision scene (#39 ) has several issues:

1. We have not yet validated that the wheelchair location relative to the robot actually aligns with the true wheelchair. To do this, we should visualize the depth map from the robot's camera in RVIZ and move the wheelchair collision object to align with the real wheelchair.
2. We have one large `wheelchair_collision_object` which is meant to expand the wheelchair to account for the user's body/. However, user's have a variety of body types. We have currently implemented a `wheelchair_collision_object` focused on a large, tall body type. However, if someone who is shorter sits in the wheelchair, their head is in inside the `wheelchair_collision_object` which means the robot cannot move to their head. We currently address this by turning off the `wheelchair_collision_object` (see #64 ) but that is dangerous. Instead, we should do one or more of the following:
    1. We should make the `wheelchair_collision_mesh` more granular, so as opposed to having one mesh that expands the wheelchair and adds the user body, we separate this into two meshes. Then, when we move the head mesh in bite transfer, we should also move the body mesh. Then we won't have to turn off the body mesh during transfer.
    2. We should rely on an Octomap to get real-time perception of the user, as opposed to relying on pre-defined meshes of abstract human bodies/heads..

amalnanavati:
We should also consider expanding the collision box around the face when not doing the final approach to or retraction from the user's mouth.

amalnanavati:
As of [ada_ros2#16](https://github.com/personalrobotics/ada_ros2/pull/16), the robot rotates with wheelchair rotations, but the collision meshes do not. We need to find a way to handle this. The simplest way would be to just divide the wheelchair and wheelchair collision mesh in half and rotate the upper half with the robot. Although we'd have to see whether MoveIt updates collision objects if the frame they are in updates.

amalnanavati:
Re. collision objects rotating with the robot, we should consider attaching some of the collision objects to the robot. This can be done with the [ApplyPlanningScene service](https://github.com/ros-planning/moveit_msgs/blob/humble/srv/ApplyPlanningScene.srv). That service message's fields `scene.robot_state.attached_collision_objects` lets us define collision objects that are attached to a robot link.

Issue #61, opened 2023-08-30
[ROS2] Make Planning or Execution Optional for MoveTo


amalnanavati:
Currently, the MoveTo behavior executes both planning and execution. However, there might be cases where we want to do multi-stage planning (e.g., plan the movement to the food and the movement within the food) before executing them both. To enable something like this, MoveTo must have the option to either plan or execute or both.

- If it is set to just plan, it should write the trajectory to blackboard.
- If it is set to just execute, it should read the trajectory from the blackboard.
- If it set to do both, it doesn't read or write the trajectory to/from the blackboard.

amalnanavati:
Addressed with #115

Issue #54, opened 2023-08-29
[ROS2] Add cleanup function to `create_action_servers.py`


amalnanavati:
Some trees make  changes (e.g., allowing certain collisions, toggling perception nodes on) that should be undone when the tree ends. The tree can end in a couple of circumstances:

1. Tree succeeds: The tree should be designed in such a way that all behaviors necessary to undo the changes are taken care of if the tree succeeds.
2. Tree fails: In this case, we can add a selector with memory to run certain cleanup behaviors after failure.
3. ROS action fails or is aborted: In this case, we need a cleanup function in `create_action_servers.py` that cleans up any state from the action.

This Issue is to implement #3.

amalnanavati:
I'd propose the following:

1. `action_server_bt.py` already has a `preempt_goal(...)` function that is called when the action is preempted or the watchdog fails. This function is intended to block until preemption is complete. We should use this function to achieve the desired functionality.
2. Currently, `preempt_goal(...)` calls the tree's `stop` method, which in turn calls every behavior's `terminate` method. This is important (e.g., to close any threads or other state the tree created in `setup`), but does not allow us to do tree-level cleanup (e.g., ensure certain ROS services are called).
3. Therefore, I propose changing `create_tree` into two functions that must be overridden, `create_main_tree` and `create_cleanup_tree` (which returns `None` if there is no cleanup to be done). `create_tree` calls both, stores the cleanup tree as an instance attribute, and returns a final tree consisting of both put together in a SelectorWithMemory. `preempt_goal` still calls `stop` on the tree that is passed to it (the final tree), but then also ticks the `cleanup tree` until completion. We will need to determine how often to tick it though.

The downside of the above solution is that it still relies on tree logic. This is nice because it makes sense that we would want to run certain behaviors in the cleanup. But the downside is that we have to tick it to completion and it introduces this other variable of how often to tick the tree.

However, the other solution I can think of is having each tree create an arbitrary cleanup function. However, then it would have to re-create the service clients that is necessary to cleanup, which seems overkill to me when the service client is already in the behavior. (Although, actually most likely the cleanup tree that is in the Fallback and the cleanup tree that is called in `preempt_goal` need to be separate, because `py_trees` doesn't allow the same behavior in multiple trees, so maybe we should; do an arbitrary cleanup function anyway because each behavior will have its own service client regardless.)

amalnanavati:
[...]

PR #72, opened 2023-09-02
Improve Bite Transfer

### LINKED ISSUE #53: [ROS2] MoveAwayFromMouth Actions
Currently, when moving away from the mouth, the web app calls the vanilla `MoveAbovePlate` and `MoveToRestingPosition`. However, the trouble with this is:
1. When near the user's mouth, the robot is in collision with the expanded `wheelchair_collision` object. Therefore, no vanilla action will work. We need an action that will allow collisions with the `wheelchair_collision` object, initially (e.g., as the robot moves back to the staging location) and then disallows collisions as it either moves above the plate or moves to the resting position.
2. Using vanilla actions prevents us from doing any other customizations of robot motion when moving away from the mouth (e.g., lowering the speed).

Note that currently MoveToMouth #42 permanently allows collisions with the `wheelchair_collision` object to account for this, which is dangerous. Therefore, addressing this is a top priority.

### LINKED ISSUE #64: [ROS2] Improve Bite Transfer
The MVP Bite Transfer in #42 and #56 results in some scary/dangerous motions, for a few reasons:

1. The movement from the staging position to the mouth is a kinematic motion that has few constraints, and can therefore be a large motion that gets out of the user's eyesight, gets very close to the user, etc.
2. We turn off the wheelchair collision object for the duration of the final movement to the mouth, which can result in trajectories where other joints of the robot get too close to or collide with the body.
3. The entire MoveFromMouth motion has the wheelchair collision constraint disabled (as opposed to just for the small part of the motion that gets out of collision with that mesh).

The MVP way to fix this is fundamentally moving back to the ROS1 bite transfer:
1. Make the motion from the staging location to the mouth be Cartesian.
2. For the MoveFromMouth actions, have it be composed of two motions, one back to the staging location (with the expanded wheelchair collision turned off), and the other to the above plate or resting position (with it turned on).
3. Potentially lower the speeds when moving towards and away from the mouth (#58 ).

As part of doing that, we should maybe move back to the center staging location for now, as opposed to the side one. We should also consider moving that staging location down, to account for people who are shorter.

Note that part of the issue is about collision constraints. We really only want to turn off collisions with an expanded mesh close to the head, but we end up turning off collisions with the entire expanded wheelchair collision object (meant to represent the user's body. That is problematic. Even if we do add a separate "wheelchair collision object" (meant to represent a user's body) and a separate "head collision object," the challenge is that people are different sizes and for some people their head will be in the wheelchair collision object. So maybe we have to get the Octomap working to deal with this?

amalnanavati:
# Description

In service of #64 , this PR improves bite transfer in the following ways:

1. Moves the staging location to in-front of the face, since that location works better for cartesian motions to approahc the mouth.
2. Makes the approach to the mouth cartesian. Adds a jump threshold (to prevent joint jumps that result in non-cartesian motions by the post-processor), and required fraction of the path that must be completed to consider it a success.
3. Creates MoveFromMouth actions, that first move to the staging location, and then move to a specified end configuration (this PR replaces #56 ).
4. Tunes the `range` parameter for RRTStar to result in plans that don't make unnecessarily large, overshooting motions. See [ada_ros2#17](https://github.com/personalrobotics/ada_ros2/pull/17).
5. Other small improvements:
    1. Moves some of the behaviors used in both MoveToMouth and MoveFromMouth into an idiom.
    2. Implements proper cleanup as a fallback of the tree.

# Testing procedure

1. Pull the latest code in the following packages:
    1. `ada_feeding`: this branch `amaln/improve_bite_transfer`
    2. `pymoveit2`: [branch `amaln/cartesian_avoid_collision`](https://github.com/personalrobotics/pymoveit2/tree/amaln/cartesian_avoid_collision). 
    3. `ada_ros2`: [branch `amaln/rrt_range`](https://github.com/personalrobotics/ada_ros2/pull/17)
    4. `forque_sensor_hardware`: [branch `main`](https://github.com/personalrobotics/forque_sensor_hardware/commit/6561c670bd47455050426f285867e39dbe3ae56d)
    5. `py_trees_ros`: [branch `amaln/service_client`](https://github.com/personalrobotics/py_trees_ros/tree/amaln/service_client)
6. Follow the[ instructions in the README]
(https://github.com/personalrobotics/ada_feeding/tree/amaln/improve_bite_transfer/ada_feeding#usage) for what code to run before you can run the action servers.

- [x] Move to the resting position and ensure it works as expected: `ros2 action send_goal /MoveToRestingPosition ada_feeding_msgs/action/MoveTo "{}" --feedback`
- [x] Move to mouth and ensure it works as expected: `ros2 action send_goal /MoveToMouth ada_feeding_msgs/action/MoveTo "{}" --feedback`
- [x] Move from mouth to resting position and ensure it works as expected: `ros2 action send_goal /MoveFromMouthToRestingPosition ada_feeding_msgs/action/MoveTo "{}" --feedback`
- [x] Move back to the mouth, and then move from mouth to above plate, and ensure it works as expected: `ros2 action send_goal /MoveFromMouthToAbovePlate ada_feeding_msgs/action/MoveTo "{}" --feedback`
- [x] Terminate the perception nodes. Move to the resting position, and then move to the mouth. Ensure the motion to the mouth fails immediately (as opposed to e.g., looping forever).

# Further Improve Plan Quality

Although this PR is enough to bring bite transfer up to MVP, the following are directions to investigate to further improve plan question.
- Further tune RRT*'s parameters and/or switch to AnytimePathShortening as documented in [ada_ros2#17](https://github.com/personalrobotics/ada_ros2/pull/17).
- Try different IK solvers (e.g., [TRAC-IK](http://docs.ros.org/en/kinetic/api/moveit_tutorials/html/doc/trac_ik/trac_ik_tutorial.html), or others documented [here](https://picknik.ai/cartesian%20planners/moveit/motion%20planning/2021/01/07/guide-to-cartesian-planners-in-moveit.html)), to see if they give better solutions, where better is defined as faster, better looking, or provides us more control (e.g., to tune the parameters and produce [more human-like IK](https://picknik.ai/cartesian%20planners/moveit/motion%20planning/2021/01/07/guide-to-cartesian-planners-in-moveit.html)).

# Before Merging
- [ ] `Squash & Merge`

### DIFF
--- a/ada_feeding/README.md
+++ b/ada_feeding/README.md
@@ -21,7 +21,7 @@ This code has been developed and tested with the Kinova JACO Gen2 Arm, on comput
 4. Launch the RealSense & Perception:
     1. Dummy nodes: `ros2 launch feeding_web_app_ros2_test feeding_web_app_dummy_nodes_launch.xml run_motion:=false run_web_bridge:=false`
     2. Real nodes: Follow the instructions in the [`ada_feeding_perception` README](https://github.com/personalrobotics/ada_feeding/blob/ros2-devel/ada_feeding_perception/README.md#usage)
-5. Run the action servers: `ros2 launch ada_feeding ada_feeding_launch.xml`
+5. Ensure the e-stop is plugged in. Run the action servers: `ros2 launch ada_feeding ada_feeding_launch.xml`
 6. Launch MoveIt2:
     1. Sim (RVIZ): `ros2 launch ada_moveit demo_feeding.launch.py sim:=mock`
     2. Real Robot: `ros2 launch ada_moveit demo_feeding.launch.py`
@@ -31,7 +31,9 @@ This code has been developed and tested with the Kinova JACO Gen2 Arm, on comput
         2. `ros2 action send_goal /AcquireFood ada_feeding_msgs/action/AcquireFood "{header: {stamp: {sec: 0, nanosec: 0}, frame_id: ''}, detected_food: {roi: {x_offset: 0, y_offset: 0, height: 0, width: 0, do_rectify: false}, mask: {header: {stamp: {sec: 0, nanosec: 0}, frame_id: ''}, format: '', data: []}, item_id: '', confidence: 0.0}}" --feedback`
         3. `ros2 action send_goal /MoveToRestingPosition ada_feeding_msgs/action/MoveTo "{}" --feedback`
         4. `ros2 action send_goal /MoveToMouth ada_feeding_msgs/action/MoveTo "{}" --feedback`
-        5. `ros2 action send_goal /MoveToStowLocation ada_feeding_msgs/action/MoveTo "{}" --feedback`
+        5. `ros2 action send_goal /MoveFromMouthToRestingPosition ada_feeding_msgs/action/MoveTo "{}" --feedback`
+        6. `ros2 action send_goal /MoveFromMouthToAbovePlate ada_feeding_msgs/action/MoveTo "{}" --feedback`
+        7. `ros2 action send_goal /MoveToStowLocation ada_feeding_msgs/action/MoveTo "{}" --feedback`
     2. Test the individual actions with the web app:
         1. Launch the web app ([instructions here](https://github.com/personalrobotics/feeding_web_interface/tree/main/feedingwebapp))
         2. Go through the web app, ensure the expected actions happen on the robot.
--- a/ada_feeding/README.md
+++ b/ada_feeding/README.md
@@ -8,7 +8,7 @@ This code has been developed and tested with the Kinova JACO Gen2 Arm, on comput
 - Install [ROS2 Humble](https://docs.ros.org/en/humble/Installation.html)
 - Install Python dependencies: `python3 -m pip install pyyaml py_trees pymongo tornado trimesh`
 - Install the code to command the real robot ([instructions here](https://github.com/personalrobotics/ada_ros2/blob/main/README.md))
-- Git clone the [PRL fork of pymoveit (branch: `amaln/allowed_collision_matrix`)](https://github.com/personalrobotics/pymoveit2) and the [PRL fork of py_trees_ros (branch: `amaln/service_client`)](https://github.com/personalrobotics/py_trees_ros/tree/amaln/service_client) into your ROS2 workspace's `src` folder.
+- Git clone the [PRL fork of pymoveit (branch: `amaln/cartesian_allow_collision`)](https://github.com/personalrobotics/pymoveit2/tree/amaln/cartesian_avoid_collision) and the [PRL fork of py_trees_ros (branch: `amaln/service_client`)](https://github.com/personalrobotics/py_trees_ros/tree/amaln/service_client) into your ROS2 workspace's `src` folder.
 - Install additional dependencies: `sudo apt install ros-humble-py-trees-ros-interfaces`.
 - Install the web app into your workspace ([instructions here](https://github.com/personalrobotics/feeding_web_interface/tree/main/feedingwebapp)).
 
--- a/ada_feeding/ada_feeding/behaviors/move_to.py
+++ b/ada_feeding/ada_feeding/behaviors/move_to.py
@@ -159,6 +159,9 @@ class MoveTo(py_trees.behaviour.Behaviour):
                 self.move_to_blackboard, "allowed_planning_time", 0.5
             )
 
+        # If the plan is cartesian, it should always avoid collisions
+        self.moveit2.cartesian_avoid_collisions = True
+
         # Reset local state variables
         self.prev_query_state = None
         self.planning_start_time = time.time()
--- a/ada_feeding/ada_feeding/behaviors/move_to.py
+++ b/ada_feeding/ada_feeding/behaviors/move_to.py
@@ -267,6 +267,17 @@ class MoveTo(py_trees.behaviour.Behaviour):
                     )
                     return py_trees.common.Status.FAILURE
 
+                # MoveIt's default cartesian interpolator doesn't respect velocity
+                # scaling, so we need to manually add that.
+                if self.cartesian and self.moveit2.max_velocity > 0.0:
+                    for point in traj.points:
+                        nsec = point.time_from_start.sec * 10.0**9
+                        nsec += point.time_from_start.nanosec
+                        nsec /= self.moveit2.max_velocity
+                        sec = int(math.floor(nsec / 10.0**9))
+                        point.time_from_start.sec = sec
+                        point.time_from_start.nanosec = int(nsec - sec * 10.0**9)
+
                 # Set the trajectory's initial distance to goal
                 self.tree_blackboard.motion_initial_distance = (
                     self.distance_to_goal.set_trajectory(traj)
--- a/ada_feeding/ada_feeding/behaviors/move_to.py
+++ b/ada_feeding/ada_feeding/behaviors/move_to.py
@@ -100,6 +100,10 @@ class MoveTo(py_trees.behaviour.Behaviour):
         self.move_to_blackboard.register_key(
             key="max_velocity_scaling_factor", access=py_trees.common.Access.READ
         )
+        # Add the ability to set acceleration scaling
+        self.move_to_blackboard.register_key(
+            key="max_acceleration_scaling_factor", access=py_trees.common.Access.READ
+        )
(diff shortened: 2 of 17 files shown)


amalnanavati:
Have tested in sim. Yet to test on real.

amalnanavati:
Tested on `t0b1`. Necessary improvements:

1. The cartesian plan service doesn't take into account velocity scaling. That must be manually added (e.g., by the MoveTo behavior).
2. I'm not happy with the fact that the motion to the staging configuration sometimes takes an unnecessarily large trajectory. Maybe we should reduce the workspace size? Or Plan multiple times and pick the shortest (Related to #61 )? I can try adding position path constraints, but my hunch is that will over-constrain OMPL's planner (as documented [here](https://github.com/personalrobotics/ada_feeding/issues/64#issuecomment-1702096321)).
3. Sometimes the cartesian motion to the mouth goes mostly to the mouth, but then does a huge swivel before stopping (although it ends in a similar position). To address this, we should do the following:
    1. See if planning multiple times addresses this. If so, we can compute `n` plans and then take the shortest. This would require #61 so MoveTo can only plan without executing.
    2. If not, see what knob addresses this (step size? Increasing the goal tolerance? the [jump thresholds in the cartesian planning message](http://docs.ros.org/en/hydro/api/moveit_msgs/html/srv/GetCartesianPath.html)? etc.). Ensure that knob reliably addresses the issue.
    4. As an ultimate safety measure, we should try to detect large swivels of the arm and truncate the trajectory at that point. This does require forward kinematics along the trajectory, but it may be necessary.

The Point 1 should be addressed in this PR. Point 2 and 3 should be addressed in this PR if the solution is something other than planning multiple times. If the solution is planning multiple times, it should wait on #61 .

amalnanavati:
Updates re. each of the issues above:

1. Has gotten addressed with the recent pushes.
2. I tried multiple approaches to address this. For documentation's sake, I'll list those that worked and those that didn't:
    1. PILZ LIN Planner: After some finnagling with the parameters we pass to the MoveTo behavior, I was able to make headway on PILZ. But then I started running into it being unable to compute the IK. Before I could look deeply into addressing that, I realized that PILZ also [gets linearly interpolated poses and computes one IK solution](https://github.com/ros-planning/moveit2/blob/84421eaeb2173a59992986f2794d4b50caa57d6e/moveit_planners/pilz_industrial_motion_planner/src/trajectory_functions.cpp#L237), so it seems *very* similar to the cartesian interpolator. The benefit of PILZ is unclear.
    2. [OMPL Objectives](https://ros-planning.github.io/moveit_tutorials/doc/ompl_interface/ompl_interface_tutorial.html): No improvements with changing objectives. The Minimax / MaximizeMin objectives could not find a plan.
    3. Workspace bounds: These [only seem to impact mobile bases](https://github.com/ros-planning/moveit_msgs/blob/humble/msg/WorkspaceParameters.msg), not manipulators.
    4. Adding more path constraints: OMPL constrained planner currently [only supports a single position/orientation path constraint](https://moveit.picknik.ai/humble/doc/how_to_guides/using_ompl_constrained_planning/ompl_constrained_planning.html), and unfortunately the alternative, rejection sampling, is too slow.
    5. Modifying [RRTStar Parameters](https://github.com/ompl/ompl/blob/fca10b4bd4840856c7a9f50d1ee2688ba77e25aa/src/ompl/geometric/planners/rrt/src/RRTstar.cpp#L59): I started with the `range` parameter, which seemed most relevant. **_This works_**. The default (if set to 0.0) is ~7.0. I found better trajectories with 3.5. 1.75 has even better trajectories, but sometimes fails within 5 or 10 secs of planning time. I think the sweet spot will be somewhere between 1.75-3.5, but it requires more trails to find.
    6. Future Work: In addition to finalizing the `range` parameter tuning, other knobs worth playing with include: [CHOMP](https://moveit.picknik.ai/humble/doc/how_to_guides/chomp_planner/chomp_planner_tutorial.html) post-processer (or STOMP, although I couldn't find it in humble), and [OMPL's  AnytimePathShortening](https://moveit.picknik.ai/humble/doc/examples/ompl_interface/ompl_interface_tutorial.html).
3. I have seen this again (rarely, but twice). I have verified that it is an issue with the plan, not the controller. The anecdotal behavior is that the fork moves to the final pose, and then there is a weird motion to adjust the configuration (even though the end pose is still the same). I need to look more into the [CartesianInterpolator](https://github.com/ros-planning/moveit2/blob/366be3e376acb9a3c928dbdc2389b5943b93b766/moveit_core/robot_state/include/moveit/robot_state/cartesian_interpolator.h#L93) to understand why this is. I have saved one of the "bad" trajectories, for introspection. 
4. Additionally, I saw a new issue where the Cartesian plan only solves 2%, but still returns and gets executed. The MoveTo behavior should instead take in a threshold `fraction completed` it uses to determine success, and return failure otherwise.

amalnanavati:
Updates re. each issue:

1. Scaling cartesian plans' velocity has already been addressed.
2. The issue of large motions from the resting position to the staging location has been addressed by setting the `range` parameter to `2.5` in [ada_ros2#17](https://github.com/personalrobotics/ada_ros2/pull/17).
3. The issue of a large swivel at the end of a cartesian motion:
    1. Problem:
        1. We did not pass a [`jump_threshold`](https://github.com/ros-planning/moveit_msgs/blob/99e6898f589d3d6929c6da68e0c3f1057866aca8/srv/GetCartesianPath.srv#L30) with the cartesian motion request, so any joint jumps between consecutive IK solutions was allowed.
        2. As a result, the last (?) IK was sometimes very different from the rest.
        3. The time optimal trajectory generator (called from [here](https://github.com/ros-planning/moveit2/blob/366be3e376acb9a3c928dbdc2389b5943b93b766/moveit_ros/move_group/src/default_capabilities/cartesian_path_service_capability.cpp#L180)), then, [smoothly interpolated the joints between this joint-space jump](https://github.com/ros-planning/moveit2/blob/366be3e376acb9a3c928dbdc2389b5943b93b766/moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp#L817).
        4. That resulted in a path that is nearly cartesian, but has a huge non-cartesian swivel at the end.
    2. Solution:
        1. Set a `jump_threshold` of `4.0`. I determined this by looking at the sample trajectory I saved with a large swivel. For the configurations without the swivel, the jump threshold was maximally 2.5, but for the configuration with the swivel, the jump threshold was ~13. To be safe (and remove smaller swivels as well), I set the `jump_threshold` to `4.0`.
        2. Incidentally, the default `max_step` that was being used, [0.25cm set in pymoveit2](https://github.com/AndrejOrsula/pymoveit2/blob/fd860ecfb6468e1c3fd6b1b5afe7b2db3ded4a75/pymoveit2/moveit2.py#L1393), was resulting in trajectories with ~100 points, even though only ~17 were unique and the rest were getting removed by the post-processor. So I also increased that to 1cm.
4. I added the ability to reject plans that have not completed at least a required fraction, which I set to `0.85` for MoveToMouth and MoveFromMouth.

Rarely, the motion from the staging configuration to the resting configuration would get unnecessarily close to the user's head. I only saw it once over ~20 trajectories. This should be addressed in #65 .

amalnanavati:
At this stage, I think the improved Bite Transfer is ready to test in-person.

amalnanavati:
Issues from in-person testing:

1. The staging->resting plan sometimes runs into the `weebo` monitor. The easiest solution to this would be adding a collision wall there. As long as we're doing it, we should also add a collision wall at the left side of the wheelchair, so the robot doesn't overshoot it. We can also add a top collision wall if we want. (Effectively, we are creating a workspace. It is unfortunate that MoveIt's workspace only applies to mobile bases, not manipulators.)
    1. Incidentally, this also made me realize that the our table STL is about twice as wide as our actual table. Of course it's not crucial for it to exactly match, but just FYI.
2. The cartesian trajectories are now very jerky because of how I did velocity scaling. I should find another way to do it.

amalnanavati:
Updates on the in-person testing issues:

1. I addressed this my adding walls to the workspace, in #82 .
2. I addressed this by fixing cartesian velocity scaling to not just update the time from start, but also update the velocity and acceleration.

Ready for in-person testing again.

amalnanavati:
Tested it and it works!

amalnanavati:
This is now thoroughly tested and proven, with and without the web app, and ready to be merged in. (#92 and #94 addressed all final bugs).

The few lingering issues necessary to bring bite transfer up to MVP are #96 , #83 , and #54 .

Issue #63, opened 2023-08-31
[ROS2] Watchdog trips on LoveLace during service calls


amalnanavati:
Sometimes on LoveLace, in the process of all the service calls in `pre_moveto_config` (i.e., toggle watchdog listener off, re-tare FT sensor, toggle watchdog listener on, and set FT thresholds) the watchdog trips because a watchdog message hasn't been received in 0.5 sec (or 0.1, depending on the parameter setting). However, the `update` function of that behavior is async, so it should not be blocking. This is something that is necessary to look into -- why sometimes watchdog messages are not received for a time period. It may be related to #62 , or to this [article on deadlock](https://docs.ros.org/en/humble/How-To-Guides/Sync-Vs-Async.html).

Note that both LoveLace and Nano are using [Cyclone DDS](https://docs.ros.org/en/humble/Installation/DDS-Implementations/Working-with-Eclipse-CycloneDDS.html)

amalnanavati:
Note that this is not an issue on `t0b1` + `nano` -- not sure what is going on with `lovelace`...

amalnanavati:
It is also worthwhile considering moving the `setup` function for the tree (in [`create_action_servers.py`](https://github.com/personalrobotics/ada_feeding/blob/a1316f614172eaf7bce0c6dbfe682bf91c90b878/ada_feeding/scripts/create_action_servers.py#L366)) to only run once when the tree is created, not at the beginning of every action call.

amalnanavati:
It is also worthwhile modifying the the service and action in `py_trees_ros` such that if the wait time is 0.0, it just checks whether the server is ready as opposed to waiting for it. (although then is there any point to that call? 🤷‍♂️)

amalnanavati:
#92 mostly addresses this, lingering issues will be documented there.

amalnanavati:
NOTE: The other thing (In addition to #92 ) that may have helped with this issue is lowering the face detection rate to 3Hz (in #91 ), because face detection eats up a lot of CPU power.

amalnanavati:
This issue has stopped occuring after #92 and #94 got merged in.

PR #81, opened 2023-09-08
Remove re-taring + adjust wait for service

amalnanavati:
# Description

This PR makes 3 changes:
1. As of this commit in `forque_sensor_hardware`, it no longer stops publishing F/T data while re-taring. It's performance does drop from 100Hz to 10Hz while retaring, but that is not enough to trip the F/T sensor. As a result, we no longer need to toggle the watchdog listener on/off when re-taring (since the watchdog waits for the F/T sensor to stop publishing for 0.5 sec before tripping.
2. Currently, `create_action_servers.py` calls a tree's `setup` method every time the action is executed, and then calls its `shutdown` method at the end. This is against the spirit of py_trees: `setup` should be called once (e.g., to set up any middleware connections), and `shutdown` should be called once at the end to counteract anything that `setup` did. This PR makes that change.
3. Additionally, it doesn't make much sense to wait for a service in the `setup` method of a behavior, because that is only called once. For most of the trees execution, it just performs `update`, which fails if the service is not ready (but doesn't wait on it to be ready). Therefore, this PR changes the default time to wait on services to 0.0.

# Testing procedure

1. Follow the [instructions in the README](https://github.com/personalrobotics/ada_feeding/blob/ros2-devel/ada_feeding/README.md#usage) to launch the code.
2. Run a few actions and verify that they succeed as expected:
    - [x] `ros2 action send_goal /MoveAbovePlate ada_feeding_msgs/action/MoveTo "{}" --feedback`
    - [x] `ros2 action send_goal /MoveToRestingPosition ada_feeding_msgs/action/MoveTo "{}" --feedback`

# Before Merging
- [ ] `Squash & Merge`

### DIFF
--- a/ada_feeding/ada_feeding/idioms/retry_call_ros_service.py
+++ b/ada_feeding/ada_feeding/idioms/retry_call_ros_service.py
@@ -25,7 +25,7 @@ def retry_call_ros_service(
     key_response: Optional[str] = None,
     response_checks: Optional[List[py_trees.common.ComparisonExpression]] = None,
     max_retries: int = 3,
-    wait_for_server_timeout_sec: float = 3.0,
+    wait_for_server_timeout_sec: float = 0.0,
     logger: Optional[logging.Logger] = None,
 ) -> py_trees.behaviour.Behaviour:
     """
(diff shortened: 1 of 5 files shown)


amalnanavati:
@taylorkf when you review it, can you review it on `t0b1`? I tested it on `lovelace` and saw some weirdness with services not receiving requests, but that may be due to #63 .

amalnanavati:
Tested on `t0b1`, it all works as expected.

PR #85, opened 2023-09-09
Add acquisition context, posthoc, and policy service

egordon:
# Description

(1) Adds the AcquisitionSchema Message, which holds an instance of a CoRL2023 action

(2) Adds the services AcquisitionSelect and AcquisitionReport.

The former is designed to take in the Mask (incl. ROI) associated with the desired food item and return the action we want to execute, the latter is used to report the result of that action (plus any posthoc context) to the acquisition policy.

(3) Add the package `ada_feeding_actionselect` to implement the above services. This incorporates: [a] a Context Adapter that converts the incoming Mask + image (retrieved via topic subscription) into a single context vector, [b] the actual policy that returns a _discrete_ action distribution (i.e. a list of actions + a parallel list of probabilities), and [c] a Posthoc Adapter that can optionally change the incoming posthoc into a new vector.

Currently, the only Policy implemented is `ConstantPolicy` that always returns a single action w/ probability 1.
Currently, the only Context/Posthoc adapter implemented is `NoContext`, which completely drops the image/haptics and just send `[0.0]` to the policy.

This PR leaves a lot of TODOs, including (in order of importance):
* Adding SPANet as a Context Adapter
* Adding HapticNet and Passthrough as Posthoc Adapters
* Adding Contextual Bandits as Policies
* Adding a policy cache for learned policies (so we don't start from scratch every time)
* Adding Data collection (i.e. record the mask, image, context vector, posthoc vector, action distribution, action taken, and loss)

# Testing procedure

* In `policies.yaml`, edit `constant.kwargs.index` with the exact action in `acquisition_library.yaml` you want to get
* Run the service (defaulting to the `constant` policy): `ros2 launch ada_feeding_actionselect ada_feeding_actionselect_launch.xml`
* In a separate terminal, call action_select: `ros2 service call /ada_feeding_actionselect/action_select ada_feeding_msgs/srv/AcquisitionSelect`. Verify that the response matches the schema action from the xml.
* In that terminal, call action_report: `ros2 service call /ada_feeding_actionselect/action_report ada_feeding_msgs/srv/AcquisitionReport`, verify the response is a success.

Unfortunately, this PR also starts with the skeleton of the AcquireFood tree, so it is imperative to make sure this doesn't break anything (the YAML should still point to the dummy MoveTo tree for AcquireFood).

* Make sure this doesn't break anything.

# Before Merging
- [ ] `Squash & Merge`


ada_feeding/ada_feeding/trees/acquire_food_tree.py
```diff
@@ -0,0 +1,92 @@
...
+        # TODO: Define full tree
+        pass
+
+    # Override result to add other elements to feedback msg
+    @override
+    def get_feedback(self, tree: py_trees.trees.BehaviourTree) -> object:
+        # Docstring copied by @override
+        feedback_msg = super().get_feedback(tree)
+
+        # TODO: add action_index, posthoc, action_select_hash
```
amalnanavati: Not necessary to address in this PR, but we expect the action_index, posthoc, and action_select_hash to be in the result, not the feedback.

ada_feeding/config/ada_watchdog.yaml
```diff
@@ -14,7 +14,7 @@ ada_watchdog:
     # If the force_torque sensor has not published a message in this amount of time,
     # or any of its dimensions are zero-variance in this amount of time, the watchdog 
     # will trigger a fault.
-    ft_timeout_sec: 0.5
+    ft_timeout_sec: 1.0
```
amalnanavati:
```suggestion
    ft_timeout_sec: 0.5
```

Undo the above change please, it was a temporary workaround to the re-taring issue that we have now addressed.

ada_feeding_actionselect/ada_feeding_actionselect/policy_service.py
```diff
@@ -0,0 +1,275 @@
...
+            self.ros_objs.append(self.create_subscription(
+                get_img_msg_type("~/image", self), "~/image", self.rgb_callback, 1
+            ))
+        if self.context_adapter.need_depth:
+            self.ros_objs.append(self.create_subscription(
+                get_img_msg_type("~/depth", self), "~/depth", self.depth_callback, 1
+            ))
+
+        self.ros_objs.append(self.create_service(
+            AcquisitionSelect, "~/action_select", self.select_callback
```
amalnanavati: Nit: I'm not a fan of the fact that the package name and the service name use inconsistent snake case. The package treats `actionselect` as one word (hence not separated by an underscore), whereas the service name treats it as two words.

I have a marginal preference towards treating it as two words (which they are), but am fine either way. Can you make it consistent (e.g., by changing the package name to `ada_feeding_action_select` and finding-and-replacing all instances of `actionselect` with `action_select`)?

ada_feeding_actionselect/ada_feeding_actionselect/adapters/__init__.py
```diff
@@ -0,0 +1,16 @@
+"""
+This package contains code that adapts incoming sensor info to context
+and posthoc context.
+"""
+
+# Generic Adapter Classes
+from .context_posthoc_adapter import ContextAdapter, PosthocAdapter
+
+# Just returns [0]
+from .context_posthoc_adapter import NoContext
```
amalnanavati: Nit: I find the name of this file confusing. `context_posthoc_adapter` makes me think it contains just the posthoc context adapter, but it actually contains the abstract classes for both ContextAdapter and Posthoc adapter.

What do you think of renaming it to something like `base_adapters.py`?

ada_feeding_actionselect/ada_feeding_actionselect/adapters/context_posthoc_adapter.py
```diff
@@ -0,0 +1,103 @@
...
+    @abstractmethod
+    def dim(self) -> int:
+        """
+        Property defining the size of the context vector
+        """
+        raise NotImplementedError("dimension not implemented")
+
+    @abstractmethod
+    def get_context(
+        self, mask: Mask, image: npt.NDArray, depth: npt.NDArray
```
amalnanavati:
```suggestion
        self, mask: Mask, image: Optional[npt.NDArray], depth: Optional[npt.NDArray]
```

You also need to import `Optional` from `typing` above

ada_feeding_actionselect/ada_feeding_actionselect/adapters/context_posthoc_adapter.py
```diff
@@ -0,0 +1,103 @@
...
+class ContextAdapter(ABC):
+    """
+    An interface to translate a visual Mask to a context vector.
+    """
+
+    def __init__(self):
+        """
+        Default self properties
+        """
+        self.need_rgb = False
```
amalnanavati: Can you add a comment here that these attributes are used by the policy service to determine whether to pass image and/or depth data into the context adapter?

I always find it confusing when a class defines attributes in `__init__` that are not used in that class; I think a comment will help understandability of this.

ada_feeding_actionselect/ada_feeding_actionselect/adapters/context_posthoc_adapter.py
```diff
@@ -0,0 +1,103 @@
...
+    """
+    An instantiation that just returns [0] for context and posthoc.
+    """
+
+    @property
+    def dim(self) -> int:
+        return 1
+
+    def get_context(
+        self, mask: Mask, image: npt.NDArray, depth: npt.NDArray
```
amalnanavati: Same comment re. `Optional`

ada_feeding_msgs/msg/AcquisitionSchema.msg
```diff
@@ -0,0 +1,59 @@
+# This message is used to specify an Acquisition Action based
+# on the schema available at:
+# https://drive.google.com/file/d/1IP74g9bTQAbf8lahd7nxqTfV1Aook3bG/view
```
amalnanavati, line 6: Not crucial, but I think it would be nice to add citation text (name, authors, venue) before this link, so users know to expect a PDF of a paper when they click on this.

ada_feeding_msgs/msg/AcquisitionSchema.msg
```diff
@@ -0,0 +1,59 @@
...
+
+### Reference Frames
+
+## Food Frame:
+# Origin: Center, Bottom of the food
+# +Z = against gravity
+# +X = major axis of food bounding ellipse
```
amalnanavati:
```suggestion
# +X = major axis of food bounding ellipse (see below for handling PI-rotation symmetry)
```

ada_feeding_msgs/msg/AcquisitionSchema.msg
```diff
@@ -0,0 +1,59 @@
...
+
+## Approach Frame
+# Same as Food Frame, except:
+# +X = -(pre_transform.translation())
+# Projected onto Food X/Y plane
```
amalnanavati, line 19: I always like pairing math with English for readability. I'd suggest adding another line after this:
```
# e.g., +X points in the direction from the utensil to the food.
```

ada_feeding_msgs/msg/AcquisitionSchema.msg
```diff
@@ -0,0 +1,59 @@
...
+# If set, choose the one closest to this orientation
+# In World Frame
+geometry_msgs/Vector3 pre_rot_hint
+
+### Grasp
+
+# In-Food Twist
+# Linear (dX, dY, dZ) in *approach frame*
+# Angular (omegaX, omegaY, omegaZ) in *utensil frame*
+geometry_msgs/Twist grasp_twist
```
amalnanavati: I'd suggest making this two separate [`Vector3Stamped`](http://docs.ros.org/en/melodic/api/geometry_msgs/html/msg/Vector3Stamped.html) instead, called `grasp_offset` and `grasp_rot` (or `grasp_linear` and `grasp_angular`). I find it un-intuitive that components of the same message have different reference frames.

This would also make the message definition align better with `acquisition_library.yaml`

ada_feeding_msgs/msg/AcquisitionSchema.msg
```diff
@@ -0,0 +1,59 @@
...
+# Grasp Force/Torque Thresholds
+float64 grasp_force
+float64 grasp_torque
+
+### Extraction
+
+# Extraction Twist
+# Linear (dX, dY, dZ) in *approach frame*
+# Angular (omegaX, omegaY, omegaZ) in *utensil frame*
+geometry_msgs/Twist ext_twist
```
amalnanavati: See above comment for two separate `Vector3Stamped` frames

ada_feeding_msgs/srv/AcquisitionSelect.srv
```diff
@@ -0,0 +1,26 @@
...
+# Size == 0 or len(actions)
+float64[] probabilities
+
+# Status Message
+string status
+
+# Selection ID
+# Used in AquisitionReport
+# Valid for the server node lifetime
+string id
```
amalnanavati, line 26: The name changes. Here, we call it a `Selection ID`, but in `AcquisitionReport` and `AcquireFood` we call it `action_select_hash`. We should be consistent. I have a marginal preference for `selection_id`, but am also okay with `action_select_hash`

ada_feeding_actionselect/ada_feeding_actionselect/helpers.py
```diff
@@ -0,0 +1,101 @@
...
+        schema.grasp_twist.linear.x = element["grasp_offset"][0]
+        schema.grasp_twist.linear.y = element["grasp_offset"][1]
+        schema.grasp_twist.linear.z = element["grasp_offset"][2]
+        schema.grasp_twist.angular.x = element["grasp_rot"][0]
+        schema.grasp_twist.angular.y = element["grasp_rot"][1]
+        schema.grasp_twist.angular.z = element["grasp_rot"][2]
+
+        schema.grasp_duration.sec = int(element["grasp_duration"])
+        decimal = element["grasp_duration"] - schema.grasp_duration.sec
+        schema.grasp_duration.nanosec = int((decimal * 1000000000) % 1000000000)
```
amalnanavati:
```suggestion
        schema.grasp_duration.nanosec = int((decimal * 10**9) % 10**9)
```

That way, readers won't have to count the number of zeros :)

ada_feeding_actionselect/ada_feeding_actionselect/helpers.py
```diff
@@ -0,0 +1,101 @@
...
+        schema.ext_twist.linear.x = element["ext_offset"][0]
+        schema.ext_twist.linear.y = element["ext_offset"][1]
+        schema.ext_twist.linear.z = element["ext_offset"][2]
+        schema.ext_twist.angular.x = element["ext_rot"][0]
+        schema.ext_twist.angular.y = element["ext_rot"][1]
+        schema.ext_twist.angular.z = element["ext_rot"][2]
+
+        schema.ext_duration.sec = int(element["ext_duration"])
+        decimal = element["ext_duration"] - schema.ext_duration.sec
+        schema.ext_duration.nanosec = int((decimal * 1000000000) % 1000000000)
```
amalnanavati:
```suggestion
        schema.ext_duration.nanosec = int((decimal * 10**9) % 10**9)
```

ada_feeding_actionselect/ada_feeding_actionselect/policies/__init__.py
```diff
@@ -0,0 +1,10 @@
+"""
+This module contains policies for selecting actions
+based on context (visual and posthoc).
+"""
+
+# Generic Policy
+from .constant_policy import Policy
```
amalnanavati: Again, it seems unintuitive to me that the abstract class policy is in the `constant_policy.py` file. Perhaps rename that file to `base_policies.py`? (Which would also align well with the suggested change of `base_adapters.py`)

ada_feeding_actionselect/ada_feeding_actionselect/policies/constant_policy.py
```diff
@@ -0,0 +1,135 @@
...
+    """
+    Execute a constant policy
+    """
+
+    def __init__(
+        self, context_dim: int, posthoc_dim: int, library: str = "", index: int = 0
+    ):
+        """
+        Default self properties
+        """
```
amalnanavati: Nit: explain parameters in docstring

ada_feeding_actionselect/ada_feeding_actionselect/helpers.py
```diff
@@ -0,0 +1,101 @@
...
+    -------
+    Array of AcquisitionSchema message objects
+    """
+
+    package_share = get_package_share_directory("ada_feeding_actionselect")
+    filename = os.path.join(package_share, library_path)
+
+    yaml_file = None
+    with open(filename, "r", encoding='utf-8') as file:
+        yaml_file = yaml.safe_load(file)
```
amalnanavati: I'd suggest putting this into a try-except and raising an informative error (e.g., `f"Failed to load acquisition library at {filename}. Are you passing in the right library parameter value? {exc}"`)

ada_feeding_msgs/msg/AcquisitionSchema.msg
```diff
@@ -0,0 +1,59 @@
...
+# +Z = against gravity
+# +X = major axis of food bounding ellipse
+
+## Approach Frame
+# Same as Food Frame, except:
+# +X = -(pre_transform.translation())
+# Projected onto Food X/Y plane
+
+## Utensil Frame = "forkTip" TF
+
```
amalnanavati, line 22: See comment in `acquisition_library.yaml` re. a unique name/id per action.

ada_feeding_actionselect/config/acquisition_library.yaml
```diff
@@ -0,0 +1,239 @@
+# Library of Acquisition Actions
+# Quats: W, X, Y, Z (matching Eigen)
+actions:
+    - # Default Action (VS90)
```
amalnanavati: Can you add a unique name/ID to each action here (and in the AcquisitionSchema message)? That'll make it easier if we ever need to check equality across actions (e.g., even if action select tells the robot to retry the same action, the robot could realize that it already tried that action and instead try the most likely unique action).

ada_feeding_actionselect/ada_feeding_actionselect/policies/constant_policy.py
```diff
@@ -0,0 +1,135 @@
...
+        ----------
+        context: Of dimension (context_dim, )
+        response: service response passed to the callback
+
+        Returns
+        -------
+        The input response populated based on the policy.
+        See AcquisitionSelect.srv
+        """
+        response.actions = [self.library[self.index]]
```
amalnanavati: We should keep in mind that when we implement the acquisition tree, it cannot make any assumptions that messages from this service will have the same actions in the same order. Each response may have an arbitrary subset of actions from the library, in an arbitrary order.

(The only way around this is to include all library actions in every response, and only change the probabilities. But that makes the messages unnecessarily large.)

ada_feeding_actionselect/ada_feeding_actionselect/policy_service.py
```diff
@@ -0,0 +1,275 @@
...
+            response.success = False
+            return response
+
+        # Run the policy
+        response.status = "Success"
+        response.success = True
+        response = self.policy.update(posthoc, request, response)
+        return response
+
+    def get_kwargs(self, kws_root: str, kwarg_root: str) -> Dict:
```
amalnanavati: Not crucial in this PR, but we should consider making this a generic helper in `ada_feeding.helpers` so we can also use it in `create_action_servers.py`.

Maybe add a TODO for it in-code?

ada_feeding_actionselect/ada_feeding_actionselect/policy_service.py
```diff
@@ -0,0 +1,275 @@
...
+        )
+
+        policy_cls_name = policy_cls_param.value
+        if policy_cls_name is None:
+            raise ValueError("Policy Class is required.")
+
+        policy_cls = import_from_string(policy_cls_name)
+        assert issubclass(policy_cls, Policy), f"{policy_cls_name} must subclass Policy"
+
+        kwargs = self.get_kwargs(f"{policy_name}.kws", f"{policy_name}.kwargs")
```
amalnanavati: Nit: can you rename this to `policy_kwargs` (and make the corresponding change when you initialize `self.policy` below?) `kwargs` is typically used for keyword arguments passed into a function e.g., `def __init__(self, **kwargs)`

ada_feeding_actionselect/ada_feeding_actionselect/policy_service.py
```diff
@@ -0,0 +1,275 @@
...
+        self.depth = None
+        self.ros_objs = []
+
+        if self.context_adapter.need_rgb:
+            self.ros_objs.append(self.create_subscription(
+                get_img_msg_type("~/image", self), "~/image", self.rgb_callback, 1
+            ))
+        if self.context_adapter.need_depth:
+            self.ros_objs.append(self.create_subscription(
+                get_img_msg_type("~/depth", self), "~/depth", self.depth_callback, 1
```
amalnanavati: Does this need to be aligned_depth? (i.e., do you envision context adapters making that assumption?) If so, please rename to `~/aligned_depth` to make that clear. (We did this in face and food detection, which needs aligned depth since it assumes the pixels line up.)

ada_feeding_actionselect/ada_feeding_actionselect/policy_service.py
```diff
@@ -0,0 +1,275 @@
...
+
+    # Services
+    def select_callback(self, request, response):
+        """
+        Implement AcquisitionSelect.srv
+        """
+
+        # Run the context adapter
+        context = self.context_adapter.get_context(
+            request.food_context, self.image, self.depth
```
amalnanavati: In general, the services and subscribers can be called from separate threads. Therefore, `self.image` and `self.depth` should be protected with `threading.Lock` instances. See [food detection](https://github.com/personalrobotics/ada_feeding/blob/60af2be5fdce0cbb7cab5897820cef28bf96d9a7/ada_feeding_perception/ada_feeding_perception/segment_from_point.py#L278) as an example.

ada_feeding_actionselect/ada_feeding_actionselect/policy_service.py
```diff
@@ -0,0 +1,275 @@
...
+        self.image = ros_msg_to_cv2_image(msg)
+
+    def depth_callback(self, msg: Union[CompressedImage, Image]):
+        """
+        Store depth image
+        """
+        self.depth = ros_msg_to_cv2_image(msg)
+
+    # Services
+    def select_callback(self, request, response):
```
amalnanavati: Nit: types in both `select_callback` and `report_callback`

ada_feeding_actionselect/launch/ada_feeding_actionselect_launch.xml
```diff
@@ -0,0 +1,12 @@
+<launch>
+  <arg name="policy" default="constant" description="Which policy to use" />
+
+  <!-- Launch the action servers necessary to move the robot -->
+  <node pkg="ada_feeding_actionselect" exec="policy_service" name="ada_feeding_actionselect" respawn="false">
+    <param from="$(find-pkg-share ada_feeding_actionselect)/config/policies.yaml"/>
+    <param name="policy" value="$(var policy)"/>
+    <remap from="~/image" to="/camera/color/image_raw/compressed"/>
+    <remap from="~/depth" to="/camera/aligned_depth_to_color/image_raw"/>
```
amalnanavati: See comment about `~/aligned_depth` in `policy_service.py`

ada_feeding_actionselect/ada_feeding_actionselect/policies/constant_policy.py
amalnanavati: (There is nothing to change here, it is just a note)

ada_feeding_actionselect/ada_feeding_actionselect/policies/constant_policy.py
```diff
@@ -0,0 +1,135 @@
...
+        response.probabilities = [1.0]
+        return response
```
amalnanavati: Should the constant policy still return an ID/hash (e.g., if we want to use it to test posthoc)?

amalnanavati (approved):
All looks good, and I re-ran the tests and they worked.

I put lots of comments, but I think the only one that **needs** to be addressed in this PR is undo-ing the `ada_watchdog.yaml` change. For all other comments, I'd suggest you read them and make a judgement call for whether to address them in this PR, create an issue/TODO to address them later, or not address them at all -- I'm fine in any of those cases.

ada_feeding_msgs/msg/AcquisitionSchema.msg
egordon: I can switch this to 2 Vector3 messages.
The problem with `Vector3Stamped` is that there is no guarantee that the approach / utensil frames exist in the TF tree, so we can't promise that.

ada_feeding_actionselect/config/acquisition_library.yaml
egordon: I think this is too likely to introduce programmer error (i.e. having the same ID for different actions). I think an action is adequately identified by its unique vector (everything except for `pre_rot_hint`).

ada_feeding_msgs/msg/AcquisitionSchema.msg
egordon: See reply below

ada_feeding_actionselect/config/acquisition_library.yaml
egordon: And in the discrete case all actions are manually defined, so no need to worry about float rounding issues.

In the slightly more complicated case where actions are programmatically defined, it is much more likely that they are programmatically defined in advance and added to a YAML file before execution, so we still won't run into any float rounding issues.

ada_feeding_actionselect/ada_feeding_actionselect/policies/constant_policy.py
egordon: Agree, I think the acquisition tree's job should just be (1) sampling from the action vector according to the probability vector, and then (2) executing the resulting action.

ada_feeding_actionselect/ada_feeding_actionselect/policies/constant_policy.py
egordon: It doesn't need to, since the main purpose of ID is to connect Report calls with the specific info returned by Select call in `update`. But since the constant policy doesn't actually *use* any info from the select call in `update`, it doesn't need the ID.

We can still test posthoc calls without the ID / with a blank ID.

Issue #67, opened 2023-08-31
[ROS2] Face Detection Accuracy Check


amalnanavati:
In MoveToMouth (as of #42 ), the robot arm often gets closer than 5cm to the detected mouth, and also oftentimes is in front of the nose rather than the mouth. The face detection img seems to be correct. Further, it does move 5cm away from the mouth of the head mesh in the MoveIt planning scene (it determines the planning scene head location based on the stomion detected from face detection). This makes me think that the stomion location detected in FaceDetection may be wrong (although it is possible that maybe the issue lies in the MoveToMouth behaviors).

This issue is focused on checking, in a very detailed manner, how well the detected mouth location aligns with the real mouth location. One way to do this can be by using [depth_image_proc to visualize the depth image as a pointcloud](https://github.com/ros-perception/image_pipeline/blob/humble/depth_image_proc/launch/point_cloud_xyz.launch.py) in RVIZ, and seeing how the actual mouth center aligns with where the robot moved the mouth center in the MoveIt planning scene. 

(Note that if you use [depth image proc, you have to comment out lines 47-49 and 72-74](https://github.com/ros-perception/image_pipeline/blob/humble/depth_image_proc/launch/point_cloud_xyz.launch.py) for the launchfile to work)

amalnanavati:
Incidentally, using depth_image_proc to check the alignment between the real head and the detected head is a great stepping stone into addressing the first todo of #65 , where we check whether the depth image of the wheelchair aligns with the object in the planning scene.

amalnanavati:
Note that if there is a camera calibration issue, that could lso be cauing this behavior. 

Also, if the aligned depth image has the noisy translations we were seeing before, and the Face Detection node happens to calculate depth on that frame, that may also have these issues.

amalnanavati:
Notes re. Face Detection while testing #72 :

1. We have fixed the aligned camera issue, so that is no longer a possible cause for a bug.
2. If the fork is overlapping the mouth, the detected mouth depth might be the fork. This is bad for a couple of reasons: 
    1. The robot won't move to the user's mouth.
    2. Because we move the head collision object to the location detected by face detection, we move the head collision object to overlap with the fork, which means that *no subsequent plans* will succeed.
[...]

PR #87, opened 2023-09-11
Changed depth perception to point around nose bridge

taylorkf:
# Description

This PR addresses issue #67. Occasionally, the fork is positioned in front of the user's mouth while face detection runs. This can lead to the camera detecting the depth of the fork rather than the user's mouth. This pull request updates the depth perception to a 9x9 square around the bridge of the nose, which does not get covered by the fork.

# Testing procedure

Test and run using the same procedure as https://github.com/personalrobotics/ada_feeding/pull/36#issue-1805648854. I have not run this code on the robot yet.

# Before Merging
- [ ] `Squash & Merge`

### DIFF
--- a/ada_feeding_perception/ada_feeding_perception/face_detection.py
+++ b/ada_feeding_perception/ada_feeding_perception/face_detection.py
@@ -123,9 +123,8 @@ class FaceDetectionNode(Node):
         # (updated with subscription)
         self.camera_matrix = [614, 0, 312, 0, 614, 223, 0, 0, 1]
 
-        # Keeps track of the detected mouth points
-        self.img_mouth_center = []
-        self.img_mouth_points = []
+        # Keeps track of the detected depth point
+        self.img_eyes_center = []
         self.is_face_detected = False
 
         # The header of the latest color image with a detected face/mouth
@@ -297,8 +296,7 @@ class FaceDetectionNode(Node):
 
         with self.mouth_detection_lock:
             is_face_detected = self.is_face_detected
-            img_mouth_center = self.img_mouth_center
-            img_mouth_points = self.img_mouth_points
+            img_eyes_center = self.img_eyes_center
             color_img_header = self.color_img_header
 
         with self.camera_info_lock:
@@ -309,8 +307,8 @@ class FaceDetectionNode(Node):
             face_detection_msg = FaceDetection()
             if is_face_detected:
                 # Retrieve the 2d location of the mouth center
-                u = int(img_mouth_center[0])
-                v = int(img_mouth_center[1])
+                u = int(img_eyes_center[0])
+                v = int(img_eyes_center[1])
 
                 # Find depth image closest in time to saved color image
                 difference_array = np.zeros((len(self.depth_buffer),), dtype=np.float32)
@@ -327,11 +325,13 @@ class FaceDetectionNode(Node):
                     desired_encoding="passthrough",
                 )
 
-                # Retrieve the depth value averaged over all mouth coordinates
+                # Retrieve the depth value averaged over a 9x9 pixel square around
+                # point between eyes
                 depth_sum = 0
-                for point in img_mouth_points:
-                    depth_sum += closest_depth[int(point[1])][int(point[0])]
-                depth = depth_sum / float(len(img_mouth_points))
+                for x in range(u - 4, u + 4):
+                    for y in range(v - 4, v + 4):
+                        depth_sum += closest_depth[int(x)][int(y)]
+                depth = depth_sum / float(81)
 
                 # Create target 3d point, with mm measurements converted to m
                 # Equations and explanation can be found at https://youtu.be/qByYk6JggQU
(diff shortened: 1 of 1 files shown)


ada_feeding_perception/ada_feeding_perception/face_detection.py
```diff
@@ -327,11 +325,13 @@ def depth_callback(self, msg: CompressedImage):
...
 
-                # Retrieve the depth value averaged over all mouth coordinates
+                # Retrieve the depth value averaged over a 9x9 pixel square around
+                # point between eyes
                 depth_sum = 0
-                for point in img_mouth_points:
-                    depth_sum += closest_depth[int(point[1])][int(point[0])]
-                depth = depth_sum / float(len(img_mouth_points))
+                for x in range(u - 4, u + 4):
+                    for y in range(v - 4, v + 4):
```
amalnanavati, line 332: Hmm, what do you think about making this `4` be configureable by parameter? To allow us to easily tune it.

ada_feeding_perception/ada_feeding_perception/face_detection.py
```diff
@@ -327,11 +325,13 @@ def depth_callback(self, msg: CompressedImage):
...
+                        depth_sum += closest_depth[int(x)][int(y)]
```
amalnanavati, line 333:
```suggestion
                        depth_sum += closest_depth[int(y)][int(x)]
```

Y and X should be swapped, right?

amalnanavati, line 333: Soo...as I was testing bite transfer, I began running into issues where the points used here are out-of-bounds of the depth image. What I realized is that the LBF Landmark detector can detect a face even if it is only partially in the image. If it does, it extrapolates where the points outside of the image are. So landmark points outputted by that detector may be out-of-bounds of the image.

Therefore, can you add a check here for whether y and x are in-bounds? Maybe keep track of the number of (x,y) pairs that are out of bounds, and if it is greater than a threshold proportion (0.5? perhaps configureable by parameter?) then it does not publish a depth (since our depth estimate would be unreliable).

ada_feeding_perception/ada_feeding_perception/face_detection.py
```diff
@@ -327,11 +325,13 @@ def depth_callback(self, msg: CompressedImage):
...
+                depth = depth_sum / float(81)
```
amalnanavati, line 334: I don't like this magic number. I know it comes from `9**2`, but don't think it is intuitive to readers. 

Given that some of the (x,y) pairs will get rejected for being out-of-bounds anyway, we can't assume that all 81 points will be in the `depth_sum`. Therefore, I'd just recommend enumerating the points in the depth_sum, and using that number to do the average.

ada_feeding_perception/ada_feeding_perception/face_detection.py
```diff
@@ -403,9 +403,9 @@ def camera_callback(self, msg: CompressedImage):
                 (x, y, w, h) = faces[largest_face[1]]
                 cv2.rectangle(image_rgb, (x, y), (x + w, y + h), (255, 0, 0), 2)
 
-                # Find stomion (mouth center) in image
-                img_mouth_center = landmarks[largest_face[1]][0][66]
-                img_mouth_points = landmarks[largest_face[1]][0][48:68]
+                # Find marker between eyes. This is used to estimate stomion depth in the
+                # case that the stomion is hidden behind the fork.
```
amalnanavati, line 407: Nit: this is not done "in the case that the stomion is hidden behind the fork", it is done in all cases :) 

Perhaps reword to "This is used as a proxy to estimate stomion depth, because the stomion is often hidden behind the fork."

amalnanavati (changes requested):
Mostly looks good, see requested changes below.

I think we should test it in-person before merging, just to verify everything makes sense.

ada_feeding_perception/ada_feeding_perception/face_detection.py
amalnanavati: Another suggestion: `int(x)` rounds down. But if the landmark point's position is e.g., `51.99`, that means the detector thought it was closer to `52` than `51`. To account for this I think we should round the float before casting it to an int.

amalnanavati:
Also, I just realized that this code returns the entire 3d point for the of the bridge of the node. It shouldn't do that -- the x and y should still be the stomion x and y, and the depth should be for the bridge of the nose.

However, I just realized a problem with the assumption that the depth of the bridge is the depth of the stomion: it depends on the camera pose relative to the face (because depth is in the camera frame). For example, if the RealSense is below the mouth, then the bridge of the nose will inherently be farther from the camera than the stomion is. The assumption only really holds true if the camera is half-way between the bridge and stomion, facing the user's face.

Thoughts @egordon ? The above is how the old code did it, but given that we'll be playing around with staging location, maybe we want a more reliable approach?

egordon:
@amalnanavati Relaying out discussion on the potential solution to this:

Outlier rejection: remove face points that are too close (i.e. right around the fork).

If there are at least a few mouth points remaining, average those to get the stommion depth.

If not: fit a plane to the rest of the face points (i.e. depth = np.dot(theta, [u; v; 1]), solve for theta), then use np.dot(theta, [u_stommion, v_stommion, 1]) as the depth for the mouth.

amalnanavati:
Closed by #130

Issue #62, opened 2023-08-30
[ROS2] Revamp Callback Groups (and/or Executors)


amalnanavati:
Currently, most of our nodes use the default callback group, which only lets one topic/service/action/timer process a callback at a time. This is a huge problem, because it slows down execution and can result in some callbacks not getting called for a while because other callbacks are called first.

We should be very intentional with what executor and callback group we are using for each node and each callback within the node. See the below resources for more details:
- https://docs.ros.org/en/humble/Concepts/Intermediate/About-Executors.html
- https://docs.ros.org/en/humble/How-To-Guides/Using-callback-groups.html

amalnanavati:
In #56 we are doing it in a jank-y way by resetting a protected access class attribute in `create_action_servers.py`. There has to be a better way to do it.

amalnanavati:
#90 and [ada_ros2#19](https://github.com/personalrobotics/ada_ros2/pull/19) should address this, by putting all callbacks into a ReentrantCallbackGroup (allowing them all to be executed in parallel).

While this should work in theory, I'll leave this issue open so someone can go back and more intentionally consider every pair of callbacks and whether they can truly be executed in parallel or not.

amalnanavati:
Note: the ReentrantCallbackGroups appear to be too much for `t0b1`'s and `lovelace`'s processing powers. I am seeing jerky motions and dropped watchdog messages on `t0b1`, which I never saw before, and dropped watchdog messages on `lovelace`. Perhaps the watchdog should have its own callback group, the MoveIt object should have its own callback group, and the rest should be mutually exclusive? Because I don't think we need any of the others in parallel.

amalnanavati:
In `forque_sensor_hardware` as well, I'm noticing `Error reading from telnet socket.` and `Error writing to telnet socket.` on `t0b1` if the sensor has been left on for > 1.5 mins. Perhaps the timer and the service should be in separate MutuallyExclusiveCallbacks, but not in a single ReentrantCallback?

amalnanavati:
#92 mostly addresses this, lingering issues will be documented there.

amalnanavati:
Closed with #92 and #94 . A lingering issue is to similarly revamp Quality of Service settings, documented in #93

PR #90, opened 2023-09-13
Put all callbacks in the same Reentrant Callback Group

amalnanavati:
# Description

In service of #62 .

As far as I can tell, our code is not at risk of any race conditions or deadlocks that would require mutually exclusive callbacks. (e.g., [calling a synchronous service from a callback is an example of a situation that may cause deadlock](https://docs.ros.org/en/humble/How-To-Guides/Using-callback-groups.html#avoiding-deadlocks).) However, our code does require callbacks to be received simultaneously, e.g., we should still receive watchdog messages while deciding whether or not to accept a goal.

Therefore, for every node in `ada_feeding` and `ada_feeding_perception`, this PR puts all the callbacks into the same ReentrantCallbackGroup, so they can be executed in parallel.

Paired with [ada_ros2#19](https://github.com/personalrobotics/ada_ros2/pull/19).

# Testing procedure

- [x] Launch the code as documented in the README. Run all actions. Ensure they behave as expected.
- [x] Launch the web app. Test SegmentAnything. Ensure it behaves as expected.

# Before Merging
- [ ] `Squash & Merge`

### DIFF
--- a/ada_feeding/ada_feeding/behaviors/move_to.py
+++ b/ada_feeding/ada_feeding/behaviors/move_to.py
@@ -15,7 +15,6 @@ from moveit_msgs.msg import MoveItErrorCodes
 import py_trees
 from pymoveit2 import MoveIt2, MoveIt2State
 from pymoveit2.robots import kinova
-from rclpy.callback_groups import ReentrantCallbackGroup
 from rclpy.node import Node
 from rclpy.qos import QoSPresetProfiles
 from sensor_msgs.msg import JointState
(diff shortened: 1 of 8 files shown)


amalnanavati:
@egordon You should consider doing something similar for `ada_feeding_action_select`

amalnanavati:
The `MoveToMouth` and `MoveFromMouth` actions in this were tested by rebasing #72 on top of this PR. All tests succeeded on `t0b1`

PR #94, opened 2023-09-16
Added Global MoveIt2 Object

### LINKED ISSUE #60: [ROS2] Global MoveIt2 Object

amalnanavati:
# Description

In service of #60 .

Previously, each behavior that needed to use the `MoveIt2` object created its own instance. This is non-ideal for two reasons: (1) unnecessary extra subscribers, service clients, etc. (2) If a programmer mistakenly puts two MoveTo behaviors into a Parallel composite, it will attempt to execute both because they are different instances.

This PR addresses that by doing the following:

1. Creates a global MoveIt2 object in the root of the blackboard, accessible via a helper function.
2. Implements a lock that the helper function returns. This lock should always be used in conjunction with the MoveIt2 object, to ensure thread-safety.
3. Implements a `ClearConstraints` decorator, to ensure that constraints from past behaviors do not bleed into the current behavior.
4. Modified the MoveTo trees to put a `ClearConstraints` decorator at the top of every `MoveTo` branch, but not lower than that (e.g., not between other constraints).

# Testing procedure

- [x] Run the code as documented in the README.
- [x] Run every action and verify that it succeeds.

# Before Merging
- [ ] `Squash & Merge`

### DIFF
--- a/ada_feeding/ada_feeding/behaviors/move_to.py
+++ b/ada_feeding/ada_feeding/behaviors/move_to.py
@@ -13,16 +13,15 @@ from typing import List, Optional
 from action_msgs.msg import GoalStatus
 from moveit_msgs.msg import MoveItErrorCodes
 import py_trees
-from pymoveit2 import MoveIt2, MoveIt2State
+from pymoveit2 import MoveIt2State
 from pymoveit2.robots import kinova
-from rclpy.callback_groups import ReentrantCallbackGroup
 from rclpy.node import Node
 from rclpy.qos import QoSPresetProfiles
 from sensor_msgs.msg import JointState
 from trajectory_msgs.msg import JointTrajectory
 
 # Local imports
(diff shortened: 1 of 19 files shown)


amalnanavati:
Tested on `lovelace`, all tests succeeded. For `MoveToMouth` and the `MoveFromMouth` actions, I tested them on the rebased `amaln/improve_bite_transfer`

PR #97, opened 2023-09-19
Make MoveTo use MoveIt2's joint state subscriber

### LINKED ISSUE #88: [ROS2] MoveTo Behavior should use MoveIt2's JointState Subscriber

amalnanavati:
# Description

In service of #88 . Earlier, `MoveTo` had its own joint state publisher, which unnecessarily repeated the work already done by `MoveIt2`'s joint state subscriber. This PR addresses that, by making `MoveTo` use the joint state from the `MoveIt2` object.

# Testing procedure

Load the code as documented in the README.

For each of the following actions, verify that the feedback is sent as expected and the action completes as expected:

- [x] `MoveAbovePlate`
- [x] `MoveToRestingPosition`
- [x] `MoveToMouth`
- [x] `MoveFromMouthToRestingPosition`

# Before Merging
- [ ] `Squash & Merge`

### DIFF
--- a/ada_feeding/ada_feeding/behaviors/move_to.py
+++ b/ada_feeding/ada_feeding/behaviors/move_to.py
@@ -7,17 +7,13 @@ logic to move the robot arm using pymoveit2.
 # Standard imports
 import math
 import time
-from typing import List, Optional
 
 # Third-party imports
 from action_msgs.msg import GoalStatus
 from moveit_msgs.msg import MoveItErrorCodes
 import py_trees
 from pymoveit2 import MoveIt2State
-from pymoveit2.robots import kinova
 from rclpy.node import Node
-from rclpy.qos import QoSPresetProfiles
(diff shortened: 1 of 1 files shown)


amalnanavati:
Tested on `t0b1`

PR #102, opened 2023-09-21
Add BlackboardBehavior and initial AcquireFood tree

egordon:
# Description

1. Adds class `behaviors.BlackboardBehavior`, which exposes a concise way to map blackboard keys to local strings.
2. Adds and hooks up the `AcquireFood` tree to the `AcquireFood` action server
3. Creates the behavior `ComputeFoodFrame` which, given the goal, outputs the food reference frame. It also (optionally) publishes this frame to the TF tree for debugging purposes.
4. Adds `ComputeFoodFrame` to the `AcquireFoodTree`
5. Uses `send_goal` to write specific goal arguments to the local blackboard.

**Proposal:**
In general, I think goal and result variables should be read from / written to the current tree blackboard namespace. This better enables subtree inclusion, where previous nodes explicitly write subtree inputs to that subtree's local blackboard and read subtree outputs from that subtree's local blackboard namespace.

For right now, I think it's okay for feedback to utilize the tree root blackboard. However, I think it would be better going forward for us to attach a FeedbackVisitor to construct it every tick instead. Feedback in theory should ideally only be dependent on:
* The currently running Behavior
* The blackboard attached to the currently running Behavior
* The previous Feedback

Note that, for subtrees, `get_feedback` will never be called, so `get_feedback`, `get_result`, `send_goal` should be free to assume that they are running on the main tree and act accordingly.

The upshot of all of this is that `tree_root_name` becomes deprecated, and all trees can treat their own `self.blackboard` as the root scope.

# Testing procedure

1. Use https://github.com/personalrobotics/feeding_web_interface/pull/89 (if not merged)
2. Run the Mock FT Sensor: `ros2 run ada_feeding dummy_ft_sensor.py`
3. Run the action servers: `ros2 launch ada_feeding ada_feeding_launch.xml use_estop:=false`
4. Run MoveIt in sim: `ros2 launch ada_moveit demo.launch.py sim:=mock`
5. Run this code with the default dummy carrot (oriented horizontally in the camera frame): `ros2 launch feeding_web_app_ros2_test feeding_dummy_acquirefood_launch.py`
6. Verify the Goal Succeeds (status=0). Open RViz and add an Axis Frame "food"  (rescale to 0.1 for visibility), verify that +Z is against gravity and +X is horizontal relative to the camera.
7. Run this code with the dummy carrot 2 (oriented ~vertically in the camera frame): `ros2 launch feeding_web_app_ros2_test feeding_dummy_acquirefood_launch.py request:=above_plate_2_carrot_request.pkl`
8. Verify the Goal Succeeds (status=0). Open RViz and add an Axis Frame "food" (rescale to 0.1 for visibility), verify that +Z is against gravity and +X is vertical relative to the camera and to the "lower right" in the camera frame.

# Before Merging
- [ ] `Squash & Merge`


ada_feeding/ada_feeding/behaviors/blackboard_behavior.py
```diff
@@ -0,0 +1,105 @@
...
+# Third-party imports
+import py_trees
+
+# Local imports
+from ada_feeding.helpers import BlackboardKey
+
+
+class BlackboardBehavior(py_trees.behaviour.Behaviour):
+    """
+    TODO
```
amalnanavati: Flagging

ada_feeding/ada_feeding/behaviors/blackboard_behavior.py
```diff
@@ -0,0 +1,105 @@
...
+        """
+        super().__init__(name=name)
+        self.blackboard = super().attach_blackboard_client(name=name, namespace=ns)
+        self.remap = {}
+        self.inputs = {}
+        self.outputs = {}
+
+    def blackboard_inputs(self, **kwargs) -> None:
+        """
+        Define and register all blackboard input keys.
```
amalnanavati, line 57: Can you add an English explanation of what this does, in addition to the pseudocode explanation before? e.g.,

```
Each key of the kwargs is a str specifying the location on the blackboard for that variable to be stored. Each value of kwargs is either `BlackboardKey`, in which case it specifies a blackboard location to remap the key to, or another type, in which case it specifies a constant value to store at that key. Note that as opposed to setting constants on the blackboard, this behavior stores it in a local dict.
```

ada_feeding/ada_feeding/behaviors/blackboard_behavior.py
amalnanavati: Please add to the class docstring the following:

1. `self.remap`, `self.inputs`, and `self.outputs` should not be overridden. I was hoping name mangling could allow us to prevent them from being overridden, but unfortunately that won't work (see [example](http://tpcg.io/_7WJQRX)).
2. blackboard values should only be accessed via `blackboard_exists`, `blackboard_get`, and `blackboard_write`, not by accessing `self.blackboard` directly.

ada_feeding/ada_feeding/behaviors/blackboard_behavior.py
```diff
@@ -0,0 +1,105 @@
...
+    def blackboard_get(self, key: str) -> Any:
+        """
+        Return a key from either the blackboard or local store.
+        """
+        if key in self.inputs:
+            return self.inputs[key]
+
+        return self.blackboard.get(self.remap[key])
+
+    def blackboard_write(self, key: str, output: Any) -> None:
```
amalnanavati: Why `blackboard_write` instead of `blackboard_set` (which would seem to match better with `blackboard_get`)?

ada_feeding/ada_feeding/behaviors/blackboard_behavior.py
```diff
@@ -0,0 +1,105 @@
...
+        for key, value in kwargs.items():
+            if isinstance(value, BlackboardKey):  # Catches None
+                self.blackboard.register_key(
+                    key=value, access=py_trees.common.Access.WRITE
+                )
+            self.outputs[key] = value
+
+    def blackboard_exists(self, key: str) -> bool:
+        """
+        Check if a key is set in the blackboard or available locally.
```
amalnanavati, line 108: Add to `blackboard_exists`, `blackboard_get`, `blackboard_write` docstrings: `Raises KeyError if the key has not been defined in blackboard_inputs/outputs`

ada_feeding/ada_feeding/behaviors/blackboard_behavior.py
amalnanavati: Or `blackboard_read` instead of `blackboard_get`. Basically I think we should either use read/write or get/set, not both.

ada_feeding/ada_feeding/behaviors/blackboard_behavior.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def blackboard_outputs(self, **kwargs) -> None:
+        """
+        Define and register all blackboard output keys.
+
+        Generally this is done in the subclass as follows:
+        <key>: BlackboardKey = <default>
+
+        Call from the subclass using:
+        super().blackboard_inputs(**{key: value for key,
```
amalnanavati: outputs, not inputs

ada_feeding/ada_feeding/trees/acquire_food_tree.py
```diff
@@ -73,9 +55,69 @@ def create_move_to_tree(
...
+            # Default world_frame = "world"
+            # Default debug_tf_frame = "food"
+        )
+        compute_food_frame.blackboard_outputs(
+            action_select_request=None,
+            food_frame=None,
+            debug_tf_publisher=BlackboardKey(
+                "debug_tf_publisher"
+            ),  # Set to None to disable
+        )
```
amalnanavati, line 89: Not crucial, but as it stands we have to do 3 function calls to fully define a `BlacboardBehavior`, which is not a common pattern in Python. I think it would be cleaner if `BlackboardBehavior.__init__` takes in `blackboard_inputs` and `blackboard_outputs` as a Dict parameters, and then calls the respective functions to set them (e.g., self.blackboard_inputs(**blackboard_inputs)).

ada_feeding/ada_feeding/behaviors/acquisition/compute_food_frame.py
```diff
@@ -5,56 +5,246 @@
...
+                                                        static broadcaster here to keep it alive
+                                                        for debugging purposes.
+        """
+        # pylint: disable=unused-argument
+        # Arguments are handled generically in base class.
+        super().blackboard_outputs(
+            **{key: value for key, value in locals().items() if key != "self"}
+        )
+
+    # pylint: disable=attribute-defined-outside-init
```
amalnanavati: I believe function-level pylint disable should be defined inside the function block. Else, it applies to the whole class.

(I haven't been consistent with this myself, and am gradually fixing it)

ada_feeding/ada_feeding/behaviors/acquisition/compute_food_frame.py
```diff
@@ -5,56 +5,246 @@
...
+    # defined in the setup / initialise functions.
+
+    def setup(self, **kwargs):
+        """
+        Middleware (i.e. TF) setup
+        """
+
+        # Create TF Tree
+        self.tf_buffer = Buffer()
+        self.tf_listener = TransformListener(
```
amalnanavati: [Unregister the listener](https://github.com/ros/geometry2/blob/bfcd60628a75519b9f59864c12e742f0e24184d6/tf2_ros/src/tf2_ros/transform_listener.py#L62) in `shutdown`. Shutdown is supposed to do the inverse of `setup`. That way, after `shutdown` is called, the `tf_listener` will cease to take up resources in the callback group.

(As a side-note, we should discuss whether `create_action_servers.py` should call `setup` and `shutdown` per action execution, or once for the life of the tree. We can discuss this in-person.)

ada_feeding/ada_feeding/behaviors/acquisition/compute_food_frame.py
```diff
@@ -5,56 +5,246 @@
...
+    def initialise(self):
+        """
+        Behavior initialization
+        """
+
+        # Construct camera intrinsics
+        # See: https://medium.com/@yasuhirachiba/converting-2d-image-coordinates-
+        #      to-3d-coordinates-using-ros-intel-realsense-d435-kinect-88621e8e733a
+        camera_info = self.blackboard_get("camera_info")
+        self.intrinsics = pyrealsense2.intrinsics()
```
amalnanavati, line 125: If this library works well for you, we should add an Issue to use it in FaceDetection as well. Currently, FaceDetection implements its own intrinsics calibration.

ada_feeding/ada_feeding/behaviors/acquisition/compute_food_frame.py
```diff
@@ -5,56 +5,246 @@
...
+        self.intrinsics.ppx = camera_info.k[2]
+        self.intrinsics.ppy = camera_info.k[5]
+        self.intrinsics.fx = camera_info.k[0]
+        self.intrinsics.fy = camera_info.k[4]
+        if camera_info.distortion_model == "plumb_bob":
+            self.intrinsics.model = pyrealsense2.distortion.brown_conrady
+        elif camera_info.distortion_model == "equidistant":
+            self.intrinsics.model = pyrealsense2.distortion.kannala_brandt4
+        else:
+            self.intrinsics.model = pyrealsense2.distortion.none
```
amalnanavati, line 140: Should we log an error here? This is unlikely to arise for our RealSense, which is "plumb_bob", but I'm wondering if we should raise an error in case others reuse our code.

ada_feeding/ada_feeding/behaviors/acquisition/compute_food_frame.py
```diff
@@ -5,56 +5,246 @@
...
+        if not self.tf_buffer.can_transform(
+            self.blackboard_get("world_frame"),
+            camera_frame,
+            rclpy.time.Time(),
+        ):
+            # Not yet, wait for it
+            # Use a Timeout decorator to determine failure.
+            return py_trees.common.Status.RUNNING
+        transform = self.tf_buffer.lookup_transform(
+            self.blackboard_get("world_frame"),
```
amalnanavati: Shouldn't we get the world frame once before the if statement, as opposed to twice? It doesn't matter much if the world frame is constant, but if this behavior is used with the world frame being written to the blackboard, we would be doing two blackboard lookups instead of one, which seems unnecessary.

ada_feeding/ada_feeding/behaviors/acquisition/compute_food_frame.py
```diff
@@ -5,56 +5,246 @@
...
+            camera_frame,
+            rclpy.time.Time(),
+        )
+
+        # Set up return objects
+        world_to_food_transform = TransformStamped()
+        world_to_food_transform.header.stamp = node.get_clock().now().to_msg()
```
amalnanavati, line 183: Should this instead be the timestamp of `transform`? (I'm fine either way, just wanted to flag it to make sure we have considered it.)

ada_feeding/ada_feeding/behaviors/acquisition/compute_food_frame.py
```diff
@@ -5,56 +5,246 @@
...
+from ada_feeding.helpers import BlackboardKey, quat_between_vectors
+from ada_feeding.behaviors import BlackboardBehavior
+from ada_feeding_perception.helpers import ros_msg_to_cv2_image
 
 
-class ComputeFoodFrame(py_trees.behaviour.Behaviour):
+class ComputeFoodFrame(BlackboardBehavior):
     """
-    TODO
+    Computes the food reference frame.
```
amalnanavati, line 32: Please define the food reference frame here, or point to the file where it is defined (I believe in one of the message definitions).

ada_feeding/ada_feeding/behaviors/blackboard_behavior.py
```diff
@@ -0,0 +1,105 @@
...
+    def blackboard_write(self, key: str, output: Any) -> None:
+        """
+        Write a key to the blackboard
+        """
+
+        if key not in self.outputs:
+            raise KeyError(f"{key} is not a registered output: {self.outputs.keys()}")
+
+        if self.outputs[key] is not None:
+            self.blackboard.set(self.outputs[key], output)
```
amalnanavati: I think we should log an error/warning if `blackboard_write` is called when the default value was None. A programmer may have just forgotten to change the default output value, and a warning can help surface that issue.

ada_feeding/ada_feeding/behaviors/acquisition/compute_food_frame.py
```diff
@@ -5,56 +5,246 @@
...
+        if len(self.blackboard_get("debug_food_frame")) > 0:
+            stb = StaticTransformBroadcaster(self.blackboard_get("ros2_node"))
+            stb.sendTransform(world_to_food_transform)
+            self.blackboard_write("debug_tf_publisher", stb)
+        self.blackboard_write("food_frame", world_to_food_transform)
+        request = AcquisitionSelect.Request()
+        request.food_context = mask
+        self.blackboard_write("action_select_request", request)
+
+        return py_trees.common.Status.SUCCESS
```
amalnanavati, line 263: In practice, once we have the transform how long does it take to do the computations? Because the tree is blocked until the computations are done. If it takes a while, we should instead make it an asynch function that does the computation, and return RUNNING until that asynch function succeeds.

ada_feeding/ada_feeding/helpers.py
```diff
@@ -35,6 +54,42 @@
 MOVE_TO_NAMESPACE_PREFIX = "move_to"
 
 
+def quat_between_vectors(vec_from: Vector3, vec_to: Vector3) -> Quaternion:
```
amalnanavati, line 59: Trusting that this is correct, I didn't check the math on this. In the future we can consider writing a unit test for this to check it.

ada_feeding/ada_feeding/trees/acquire_food_tree.py
```diff
@@ -73,9 +55,69 @@ def create_move_to_tree(
         -------
         tree: The behavior tree that acquires food
         """
+        # TODO: remove name and tree_root_name
+        # We have access to them implicitly via self.blackboard / self.blackboard_tree_root
```
amalnanavati, line 59: Good point, feel free to create an issue for this.

Currently, [`self.blackboard.namespace` and `self.blackboard_tree_root.namespace` are the ones that have the unchanged name](https://github.com/personalrobotics/ada_feeding/blob/74dab3aceba646fa5c770acda8c64ce03120bb80/ada_feeding/ada_feeding/trees/move_to_tree.py#L71). We should consider whether it should instead be `.name`.

amalnanavati (approved):
LGTM, comments should be addressed, but nothing that requires a re-review.

Re. your idea of writing goals in local namespace as opposed to the tree root namespace, I think the only scenario in which that matters is if we use a tree within another tree, and I'm not sure how we would percolate the goal down into every subtree's blackboard. We should discuss more asynch (e.g., on a GitHub issue). For Feedback and Result, I agree that we can create a visitor for those.

ada_feeding/ada_feeding/trees/acquire_food_tree.py
egordon: Adding dictionaries to __init__ introduces a few of its own problems:

1. It kills per-key static type-checking.
2. It allows a client of the behavior to add read/write keys to the Behavior blackboard without explicit specification by the programmer of the behavior

ada_feeding/ada_feeding/behaviors/acquisition/compute_food_frame.py
egordon: That's fair, but I think it is conceptually fine for this to be the current time since `transform` could still be in the past, and the world-to-food frame isn't valid until it is calculated.

egordon:
> Re. your idea of writing goals in local namespace as opposed to the tree root namespace, I think the only scenario in which that matters is if we use a tree within another tree, and I'm not sure how we would percolate the goal down into every subtree's blackboard. We should discuss more asynch (e.g., on a GitHub issue). For Feedback and Result, I agree that we can create a visitor for those.

We can discuss offline, though I want to flag that we shouldn't conceive of the input of the tree being the _action goal_ specifically. Instead, we can consider a tree to have a set of inputs, like a function, that are implemented as expected variables written to the local blackboard prior to tree execution.

For the root tree, `send_goal` takes the goal object and writes those variables.

For sub-trees, it is up to the parent tree to write to those variables (by having the keys be the outputs of other nodes or written directly).

ada_feeding/ada_feeding/trees/acquire_food_tree.py
amalnanavati: I don’t think it does. I’m proposing still having the functions `blackboard_inputs` and `blackboard_outputs` (to allow static type checking and to prevent programmers from adding undefined keys), but we merely provide a shortcut to initializing them via the __init__ function. Static type checking and ensuring that the right keys are passed will still be done in the `blsckboard_input`/output function

ada_feeding/ada_feeding/behaviors/blackboard_behavior.py
egordon: Switch to `blackboard_set`, using get/set to match the underlying API

ada_feeding/ada_feeding/behaviors/acquisition/compute_food_frame.py
egordon: Switching to global TF Listener / Buffer Object w/ default Node Callback group

egordon: As discussed

ada_feeding/ada_feeding/behaviors/acquisition/compute_food_frame.py
egordon: I added a warning.

ada_feeding/ada_feeding/behaviors/acquisition/compute_food_frame.py
egordon: I will say a blackboard lookup is just a hash table, so doesn't take any longer than just looking up the remapped constant value.

That said, it is probably safer in parallel trees to just do the single lookup.

ada_feeding/ada_feeding/behaviors/blackboard_behavior.py
egordon: It's expected behavior, so I'll instead write it to debug()

ada_feeding/ada_feeding/behaviors/acquisition/compute_food_frame.py
egordon: Timed: 0.0005796703s (i.e. 580us)
I think we're fine

ada_feeding/ada_feeding/helpers.py
egordon: Added issue #110

ada_feeding/ada_feeding/trees/acquire_food_tree.py
egordon: Made issue #111

Issue #101, opened 2023-09-21
[ROS2] `create_action_servers.py` should set all behaviors' logger to the node's logger


amalnanavati:
Currently, the `create_move_to_tree` function is in charge of setting every behaviors logger to the node's logger. But it is easy to forget to do so, and if we use an idiom, it becomes unwieldy to do so for all children of the idiom.

Instead, `create_action_servers.py` should iterate over the tree as part of setting it up and set every behaviors logger to the node's logger. This also makes sense because `create_action_servers.py` is the layer where the ROS connection necessarily comes in, not lower layers.

PR #103, opened 2023-09-21
Unified cleanup using scoped behaviors

amalnanavati:
# Description

In service of #54 . 

In our trees, there are numerous times when we want to flip a switch, move the robot, and then flip the switch. For example, turning face detection on/off, adding/removing a collision object in the world, allowing/disallowing collisions with certain objects, etc.

What is challenging is that we want to run the "cleanup" for flipping the switch in three different cases:

1. If the robot motion succeeds.
2. If the robot motion fails.
3. If the entire tree is preempted (e.g., the user terminates the robot motion from the app).

Achieving the above 2 is straightforward using Selectors. Achieving the third is more difficult, because when `tree.stop(INVALID)` is called, it is called on all the behaviors one-by-one, which in turn calls each behaviors' `terminate` function. Therefore, achieving the proper cleanup behavior in the third case requires performing that cleanup in a `terminate` function of a behavior.

Building off of the discussion and contribution in [`py_trees`#427](https://github.com/splintered-reality/py_trees/pull/427), this PR implements a `scoped_behavior` idiom that executes a pre-behavior, main-behavior, and post-behaviour, with a guarantee that the post-behavior will execute even if the behavior is preempted.

# Testing procedure

Run the necessary nodes as documented in the README. Verify the following:

- [x] Sim:
    - [x] `ros2 action send_goal /MoveToRestingPosition ada_feeding_msgs/action/MoveTo "{}" --feedback`
        - [x] Runs to completion successfully.
    - [x] `ros2 action send_goal /MoveToMouth ada_feeding_msgs/action/MoveTo "{}" --feedback`
        - [x] Runs to completion successfully.
        - [x] When the CLI action call is terminated while face detection is on, it toggles face detection off.
        - [x] When the CLI action call is terminated during the cartesian motion to the mouth, the robot immediately stops moving and re-enables collisions with the `wheelchair_collision` object.
    - [x] `ros2 action send_goal /MoveFromMouthToRestingPosition ada_feeding_msgs/action/MoveTo "{}" --feedback`
        - [x] Runs to completion successfully.
        - [x] When the CLI action call is terminated during the cartesian motion away from the mouth, the robot immediately stops moving and re-enables collisions with the `wheelchair_collision` object.
        - [x] When the CLI action call is terminated during the motion to the end configuration, the robot immediately stops moving and removes the `in_front_of_wheelchair_wall` from the collision scene.
    - [x] (Note that you'll have to run `MoveToMouth` and `MoveFromMouthToRestingPosition` several times iteratively to test all of the above.)
- [x] Real:
    - [x] `ros2 action send_goal /MoveToRestingPosition ada_feeding_msgs/action/MoveTo "{}" --feedback`
        - [x] Runs to completion successfully.
    - [x] `ros2 action send_goal /MoveToMouth ada_feeding_msgs/action/MoveTo "{}" --feedback`
        - [x] Runs to completion successfully.
        - [x] When the CLI action call is terminated while face detection is on, it toggles face detection off.
        - [x] When the CLI action call is terminated during the cartesian motion to the mouth, the robot immediately stops moving and re-enables collisions with the `wheelchair_collision` object.
    - [x] `ros2 action send_goal /MoveFromMouthToRestingPosition ada_feeding_msgs/action/MoveTo "{}" --feedback`
        - [x] Runs to completion successfully.
        - [x] When the CLI action call is terminated during the cartesian motion away from the mouth, the robot immediately stops moving and re-enables collisions with the `wheelchair_collision` object.
        - [x] When the CLI action call is terminated during the motion to the end configuration, the robot immediately stops moving and removes the `in_front_of_wheelchair_wall` from the collision scene.
    - [x] (Note that you'll have to run `MoveToMouth` and `MoveFromMouthToRestingPosition` several times iteratively to test all of the above.)
- [x] Run the unit tests and verify that they succeed: `colcon test --packages-select ada_feeding; colcon test-result --all --verbose`

# Before Merging
- [ ] `Squash & Merge`


amalnanavati:
Tested in sim. The only issue I noticed is that when terminating the CLI action call during robot arm motion, the tree runs the post-behavior before terminating the movement, which can be particularly harmful if the post-behavior blocks (e.g., toggling a collision object off). Instead, we want to first terminate the robot movement, and then run the post behavior.

I should write comprehensive test cases for this idiom, covering:
1. `worker` is running: behavior should be running, and `on_success`, `on_failure`, and `on_preempt` are not ticked.
2. `worker` succeeds and `on_succeed` is running: behavior should be running, and `on_failure` and `on_preempt` are not ticked.
3. `worker` succeeds and `on_succeed` is succeeds: behavior should succeed, and `on_failure` and `on_preempt` are not ticked.
4. `worker` succeeds and `on_succeed` fails: behavior should fail, and `on_failure` and `on_preempt` are not ticked.
5. `worker` fails and `on_failure` is running: behavior should be running, and `on_succeed` and `on_preempt` are not ticked.
6. `worker` fails and `on_failure` succeeds: behavior should fail, and `on_succeed` and `on_preempt` are not ticked.
7. `worker` fails and `on_failure` fails: behavior should fail, and `on_succeed` and `on_preempt` are not ticked.
8. tree is stopped before any ticks: `worker`, `on_success`, `on_failure`, and `on_preempt` are not ticked.
9. tree is stopped after a single tick where `worker` is running, fails, or succeeds: `worker` is ticked once, `on_success` and `on_failure` are not ticked, and `on_preempt` is ticked to completion.
10. tree is stopped after multiple ticks where `worker` fails/succeeds and `on_success`/`on_failure` is running: `worker` is ticked, `on_success`/`on_failure` is ticked, and `on_preempt` is ticked to completion.

amalnanavati:
Wrote comprehensive test cases, and modified `eventually_swiss` to satisfy those test cases. Simply put, the test cases guarantee that `eventually_swiss` behaves as follows:

1. `on_success` is run if and only if `workers` succeed.
2. `on_failure` is run if and only if `workers` fail
3. `on_preempt` is run if and only if a node above the root calls `root.stop(INVALID)`.
4. If `workers` succeed, the return status is the success/failure status of `on_success`.
5. If `workers` fail, the return status is failure.
6. If a node above the root calls `root.stop(INVALID)`, first `terminate(INVALID)` is called on `workers`, `on_success`/`on_failure`, and *THEN* `on_preempt` is ticked to completion.

The tree has become a bit unwieldy (reproduced below), but the comprehensive test cases mean that we can be fairly certain it works as outlined above.

```
SelectorWithoutMemory
  |-ForceFailure
  |   |-SelectorWithMemory
  |   |    |-SequenceWithMemory
  |   |    |    |-UnsetBlackboardVariable
  |   |    |    |-worker
  |   |    |    |-StatusToBlackboard
  |   |    |    |    |-on_success
  |   |    |-SequenceWithMemory
  |   |    |    |-Inverter
  |   |    |    |    |-CheckBlackboardVariableExists
  |   |    |    |-StatusToBlackboard
  |   |    |    |    |-SuccessIsFailure
  |   |    |    |    |    |-on_failure
  |-OnTerminateMultiTick
  |   |-on_preempt
  |-SequenceWithMemory
  |   |-WaitForBlackboardVariable
  |   |-BlackboardToStatus
```

All the test cases in `sim` pass. Yet to test on the real robot.

ada_feeding/ada_feeding/behaviors/move_to.py
```diff
@@ -395,7 +395,7 @@ def terminate(self, new_status: py_trees.common.Status) -> None:
         # A termination request has not succeeded until the MoveIt2 action server is IDLE
         with self.moveit2_lock:
             while self.moveit2.query_state() != MoveIt2State.IDLE:
-                self.node.logger.info(
+                self.logger.info(
```
egordon, line 398: Mental shout-out to https://github.com/personalrobotics/ada_feeding/issues/101

(I was confused where the Behavior logger was coming in)

ada_feeding/ada_feeding/decorators/force_status.py
```diff
@@ -0,0 +1,68 @@
...
+        self.logger.debug(f"{self.__class__.__name__}.tick()")
+        # initialise just like other behaviours/composites
+        if self.status != common.Status.RUNNING:
+            self.initialise()
+        # interrupt proceedings and process the child node
+        # (including any children it may have as well)
+        for node in self.decorated.tick():
+            yield node
+        # resume normal proceedings for a Behaviour's tick
+        self.status = self.update()
```
egordon: I'm confused, why do you need to override the default decorator behavior here?
Does this node switch RUNNING to const_status? I don't think that's an intuitive functionality.

My idea of ForceStatus is:
(1) const_status can't be RUNNING
(2) The status it returns is RUNNING iff decorated is RUNNING and const_status otherwise.

Also  this by-passes the normal stop/terminate behavior of children, doesn't it? That seems like it could cause unintended consequences.

ada_feeding/ada_feeding/decorators/on_preempt.py
```diff
@@ -0,0 +1,109 @@
...
+"""
+
+import time
+import typing
+
+from py_trees import behaviour, common
+from py_trees.decorators import Decorator
+
+
+class OnPreempt(Decorator):
```
egordon, line 15: To clarify, when a tree is preempted, `Behavior.stop(INVALID)` is called on *every* node in the tree (or, more accurately, it is called on the parent node, which for all Decorators/Composites calls stop(INVALID) on all children that are not already INVALID).

There's some potentially unintended behavior here.

(1) You address this below, but of course, this node needs to be ticked at least once to "activate". It is best to document that here.

(2) Once this node is activated (i.e. will catch preemptions), it appears to active *forever*, causing the preemption functionality to bleed into the entire tree. This wouldn't be caught in a unit test that only considers "local" preemptions.


Here is how I would fix this: Make this a true decorator (with the preemption subtree not a true "child") and have the preemption protection active only on the subtree.

Stopping the review here as this affects the rest of the code.

egordon (changes requested):
See comment for `OnPreempt`, stopping the review there for now.

ada_feeding/ada_feeding/decorators/force_status.py
amalnanavati: There are two reasons for this.

1. By default, [if a decorator returns a status other than RUNNING, it will call `stop(INVALID)` on all its children that are RUNNING](https://github.com/splintered-reality/py_trees/blob/0d5b39f2f6333c504406d8a63052c456c6bd1ce5/py_trees/decorators.py#L140). So therefore, we can't have a decorator that returns something like FAILURE while still allowing its kids to be ticked (as is the desired behavior here).
2. Why is it desired behavior to have this decorator return something like FAILURE while still allowing its kids to be ticked? That has to do with how composites progress in `py_trees`.
    1. We need the "main" subtree (the subtree that contains `on_success` and `on_failure`) to exist *before* the subtree containing `on_preempt`, in order to get the right preemption order. 
    2. However, both [Selectors](https://github.com/splintered-reality/py_trees/blob/0d5b39f2f6333c504406d8a63052c456c6bd1ce5/py_trees/composites.py#L455) and [Sequences](https://github.com/splintered-reality/py_trees/blob/0d5b39f2f6333c504406d8a63052c456c6bd1ce5/py_trees/composites.py#L571C23-L571C23), regardless of whether they have memory or not, don't tick later children if earlier children return `RUNNING`. Therefore, if `ForceStatus` returns `RUNNING`, then the `on_preempt` subtree will not get ticked.
    3. Why is it important to for the `on_preempt` subtree to get ticked? Because throughout the `py_trees` library, `stop(INVALID)` is only called on children who have a status other than `INVALID`. Therefore, if the `on_preempt` never gets ticked while the "main" tree is running, then the preemption behavior won't run when the tree gets preempted.
    4. So that means that if our top-level node is a `Selector` we must use `ForceStatus(FAILURE)` to ensure the `on_preempt` subtree gets ticked, and if our top-level node is a `Sequence` we must use `ForceStatus(SUCCESS)` to ensure the `on_preempt` subtree gets ticked. Between the two, `Selector` is better because it [does not call `stop(INVALID)` on children on initialization](https://github.com/splintered-reality/py_trees/blob/0d5b39f2f6333c504406d8a63052c456c6bd1ce5/py_trees/composites.py#L409) (whereas [Sequence does](https://github.com/splintered-reality/py_trees/blob/0d5b39f2f6333c504406d8a63052c456c6bd1ce5/py_trees/composites.py#L540)), which would cause `on_preempt` to run.
    5. The final option for the top-level composite to be `Parallel`, because a `Parallel` is the only composite that can have multiple children that are `RUNNING` at once. The main challenge with a `Parallel` is that once it reaches a terminal status, [it will call `stop(INVALID)`  on all its `RUNNING` children](https://github.com/splintered-reality/py_trees/blob/0d5b39f2f6333c504406d8a63052c456c6bd1ce5/py_trees/composites.py#L778), which means that the `on_preempt` behavior will be run even if the "main" tree succeeds, which is undesirable.
        1. This brings up a final point about why, at least for `scoped_behavior`, we don't just use a Parallel with `OnPreempt` (e.g., [as `py_trees` does with `eventually`](https://github.com/splintered-reality/py_trees/pull/427)). While that will achieve the desired performance of `Post` being run on succeess, failure, or preemption, it will *always run in synchronously in `terminate`*. However, desired behavior is for it to run asynchronously in `update` on success/failure, and only synchronously in `terminate` on preemption.

I think that fully covers the nuances of `py_trees` that resulted in the final architecture for `eventually_swiss`/`scoped_behavior`.

ada_feeding/ada_feeding/decorators/on_preempt.py
amalnanavati: Good point re. the problem. I'll add a unit test for that: putting this idiom into each type of composite, and verifying that once the idiom succeeds/fails and *then* the tree is terminated, the `on_preempt` behavior is *not* called.

But I don't understand your proposed fix. Let's discuss more in lab today.

ada_feeding/ada_feeding/decorators/force_status.py
egordon: Let's talk about OnPreempt, since it seems like the non-BT-esque behavior of ForceStatus is designed to cater to the weird behavior of that.

amalnanavati:
Note that even the new `scoped_behavior` implementation is not exactly what we want. Say the worker succeeds and `post` starts ticking, and then the tree is preempted. Our current implementation ticks a completely new copy of `post` to completion. In actuality, we want to continue ticking that same copy of `post`. (I have an idea for how to address this, will work on it later).

The tests that still need to be implemented for `scoped_behavior` are:

1. Preemptions: That it pre-empts correctly when preempted before, during, and after execution.
2. Nested Scope: That it behaves correctly, including with preempts, when we have nested scopes.

amalnanavati:
Addressed all changes, added comprehensive unit tests for `scoped_behavior` including nested scope. Also added unit tests for both `eventually_swiss` and `scoped_behavior` for if the tree is preempted before the idiom is ticked, and if the tree is preempted after the idiom has terminated.

@egordon note that while our discussed approach worked for `eventually_swiss`, it didn't work for `scoped_behavior`. That's because using `eventually_swiss` as a sub-idiom for `scoped_behavior` would require multiple copies of `post`, whereas we want only one copy (so that e.g., if `post` starts ticking before preemption, the same behavior will continue to get ticked). I implemented a slightly different structure for `scoped_behavior`:

```
OnPreempt(same `post` as below)
  |-SequenceWithMemory
  |   |-UnsetBlackboardVariable
  |   |-FailureIsSuccess
  |   |    |-StatusToBlackboard
  |   |    |    |-SequenceWithMemory
  |   |    |    |    |-pre
  |   |    |    |    |-workers[1]
  |   |    |    |    |-...
  |   |    |    |    |-workers[n]
  |   |-FailureIsSuccess
  |   |    |-post
  |   |-BlackboardToStatus
```

The only danger of the above formulation is potential name clashes with the blackboard status. However, I nest the blackboard variable within the idiom name, so the only way names will clash is if the programmer uses multiple idioms with the same name and doesn't pass a custom `status_blackboard_key` to it, which seems unlikely. Regardless, if you're concerned about name clashes, I can de-dup it by appending the timestamp to it. LMK.

PR #104, opened 2023-09-23
Categorical NB Food on Fork

atharva-kashyap:
# Description

This PR is adding a feature node Food on Fork that will allow for automatically detecting the presence of food on the fork. It is addressing the issue linked https://github.com/personalrobotics/ada_feeding/issues/43. It will output a confidence to a topic `/food_on_fork` and it represents the confidence with which the model is able to predict food on the fork.

## Design Decisions
Outlined, in the [readme](https://github.com/personalrobotics/ada_feeding/blob/becbbcaeaccc79bf669b19ee91d24a44688eaddb/ada_feeding_perception/README.md#food-on-fork), there is an overview of the models being used.

# Testing procedure

## Test the Training Procedure
- Make sure to download the necessary testing files (ideally store them directly inside `/ada_feeding_perception/ada_feeding_perception`). Don't unzip the files. The code will unzip and process at runtime. The three datasets as linked below:
  - [Dataset1](https://drive.google.com/file/d/1KuZonyrz4440pHjgPvTqeinP3aju0xTK/view?usp=sharing)
  - [Dataset2](https://drive.google.com/file/d/1XEIx9CipqyAJqEuqukIb9eiTW1ee8Nyu/view?usp=sharing)
  - [Dataset3](https://drive.google.com/file/d/16Gbn360WE5RXOgbfESxNhWbb9SsEgdC7/view?usp=sharing)
- Change directory into `/ada_feeding_perception/ada_feeding_perception/`. 
- Run the `food_on_fork_categorical_naive_bayes_training.py`. 
  - This file takes in a few arguments. So, follow the command outlined below:
    - General format of the command: `python food_on_fork_categorical_naive_bayes_training.py "<boolean>" "<dataset_zip>" --model_save_file "<name>"`
      - The first argument (`boolean`) specifies whether or not to use the entire dataset for training. It also specifies whether or not a model should be saved.
      - Second argument (`dataset_zip`) specifies the location of the zip file that you have downloaded from the first step. _Ideally, you should just store the downloaded zip file in `ada_feeding_perception/ada_feeding_perception`_. 
      - Third argument (`name`) is optional and is dependent on whether or not the first argument is true or false. If the first argument is true, the third argument is required, otherwise it is not required. This is the name that you want to provide for the model being saved
    - `python food_on_fork_categorical_naive_bayes_training.py "False" "./train_test_data_no_hand_8-30-23.zip"`
  - `<dataset_zip>` specifies the dataset you downloaded.
  - After running Dataset1, the expected output is as follows (confirm that this is the output you receive):
```
Train accuracy 0.999795605518651
Train confusion matrix
 [[4677    2]
 [   0 5106]]
Test accuracy 1.0
Test confusion matrix
 [[1253    0]
 [   0 1201]]
```
  - After running Datset2, the expected output is as follows:
```
Train accuracy 0.9998925193465176
Train confusion matrix
 [[4555    1]
 [   0 4748]]
Test accuracy 0.9996592844974447
Test confusion matrix
 [[1375    1]
 [   0 1559]]
```
  - After running Dataset3, the expected output is as follows:
```
Train accuracy 0.9997924234561495
Train confusion matrix
 [[4559    2]
 [   0 5074]]
Test accuracy 1.0
Test confusion matrix
 [[1371    0]
 [   0 1233]]
```
> Note that parts of this procedure is also outlined [here](https://github.com/personalrobotics/ada_feeding/blob/becbbcaeaccc79bf669b19ee91d24a44688eaddb/ada_feeding_perception/README.md#food-on-fork).

## Test the model on the robot
Now that you have confirmed that you are able to mimic the training procedure, perform a model training on the full dataset. Run the following command to train on the full dataset: `python food_on_fork_categorical_naive_bayes_training.py "True" "./train_test_data_no_hand_8-30-23.zip" --model_save_file "categorical_naive_bayes_without_hand_8-30-23.pkl"`. This should create a .pkl file, which is the model. Now either move this model into `/ada_feeding_perception/model/` folder or use the model already present inside that folder. To be able to use this model, navigate to `food_on_fork.yaml` file and change the location of the model. 

To test this model, follow these steps:
- Position the robot such that the arm is in "resting position". Make sure to listen to `/food_on_fork` topic (`ros2 topic echo /food_on_fork`). If there is food on fork, you should notice values close to 1.0 and if there is no food on fork, you should notice values close to 0.0. 
- Run the FoF node and echo `ros2 topic echo /food_on_fork` and then visually test out by adding various food items onto the fork.
- Then, reposition the arm into the "in-front of face" position, sit in the chair, and conduct the test outlined in Step 2. 
- Note: Try different food items. Try to test the extremes of adding food items as well. But, please don't get stuck on minor cases. There will still be cases where if the food is only partially on the fork and the prediction might be incorrect. 

# Before Merging
- [ ] `Squash & Merge`


ada_feeding_perception/README.md
```diff
@@ -98,3 +101,31 @@ Launch the web app along with all the other nodes (real or dummy) as documented
 - `offline.images` (list of strings, required): The paths, relative to `install/ada_feeding_perception/share/ada_feeding_perception`, to the images to test.
 - `offline.point_xs` (list of ints, required): The x-coordinates of the seed points. Must be the same length as `offline.images`.
 - `offline.point_ys` (list of ints, required): The y-coordinates of the seed points. Must be the same length as `offline.images`.
+
+## Food on Fork
+There are two models that can be used to determine Food on Fork. They are as follows:
+- Logistic Regression
+  - This model uses a single feature (number of pixels within the specified frustum). Based on that, it outputs a confidence on whether or not there is a presence of food on fork.
```
amalnanavati, line 116: Document the strengths and weaknesses of each model, and in what scenario we would want to use each.

ada_feeding_perception/README.md
```diff
@@ -98,3 +101,31 @@ Launch the web app along with all the other nodes (real or dummy) as documented
...
+  - Make sure to simply download the dataset into the folder `ada_feeding_perception/ada_feeding_perception`. Please don't unzip the file because the code unzips it at runtime.
+- To train CategoricalNB, there are a few required arguments:
+  - Suppose you want to use 80/20 train-test split and check the accuracy of the model:
+    - `python food_on_fork_categorical_naive_bayes_training.py "False" "./train_test_data_no_hand_8-30-23.zip"`:
+      - Make sure to specify the correct zipfile and the boolean argument will be false since we want to only train on the trainset and check its accuracy on the testset.
+  - Suppose you want to train and generate a `.pkl` model:
+    - `python food_on_fork_categorical_naive_bayes_training.py "True" "./train_test_data_no_hand_8-30-23.zip" --model_save_file "categorical_naive_bayes_without_hand_8-30-23.pkl"`:
```
amalnanavati: Can you make the first two arguments named arguments? That will make this command line (and the one above) easier to interpret and remember

ada_feeding_perception/README.md
```diff
@@ -98,3 +101,31 @@ Launch the web app along with all the other nodes (real or dummy) as documented
 - `offline.images` (list of strings, required): The paths, relative to `install/ada_feeding_perception/share/ada_feeding_perception`, to the images to test.
 - `offline.point_xs` (list of ints, required): The x-coordinates of the seed points. Must be the same length as `offline.images`.
 - `offline.point_ys` (list of ints, required): The y-coordinates of the seed points. Must be the same length as `offline.images`.
+
+## Food on Fork
```
amalnanavati, line 105: This is very thorough! What it is missing is instructions on actually running the FoF code to use within the system. I think it would be great to add a sentence in the intro paragraph of this section, mentioning that the node is launched by default when following the above instructions, it is always on and publishes to `topic`, and it is used by the web app to determine whether to move on from particular states.

ada_feeding_perception/ada_feeding_perception/food_on_fork.py
```diff
@@ -0,0 +1,363 @@
...
+        self.test = test.value
+        self.get_logger().info(str(self.test))
+        # self.get_logger().info(str(self.left_top_corner))
+        # self.get_logger().info(str(self.right_bottom_corner))
+        # self.get_logger().info(str(self.min_dist))
+        # self.get_logger().info(str(self.max_dist))
+
+        # depth topic subscription
+        self.subscription_depth = self.create_subscription(
+            Image, "camera/depth/image_rect_raw", self.listener_callback_depth, 1
```
amalnanavati: Following the examples elsewhere in `ada_feeding_perception`, please make this a relative topic and remap it from the launch file. See [here](https://github.com/personalrobotics/ada_feeding/blob/fb19d559d3ca72d55a0f8fc55307f5061d4e98cd/ada_feeding_perception/ada_feeding_perception/face_detection.py#L144) and [here](https://github.com/personalrobotics/ada_feeding/blob/fb19d559d3ca72d55a0f8fc55307f5061d4e98cd/ada_feeding_perception/launch/ada_feeding_perception.launch.py#L83) for examples.

amalnanavati: Additionally, this node will have to switch to the aligned depth topic. That is because the other nodes in `ada_feeding_perception` use the color image and the aligned_depth image, and if we start subscribing to 3 RealSense topics at once, then the RealSense stops reliably publishing some topics. So all nodes can only use camera and aligned_depth.

This will involve:
1. Shifting over to the aligned depth stream (again, look at face detection and the launchfile for an example), and updating the parameters for the frustrum accordingly.
2. Testing it on the real robot. In the best case, your method is robust enough that even though the perspective have shiften slightly, it still works well. In the worst case, we may have to collect a new dataset with aligned depth images and re-train the model.
3. If it turns out your existing model is still good enough even after switching to aligned depth, still add a TODO in the README that we should retrain a new model on aligned_depth, so we can do it in the future.

ada_feeding_perception/ada_feeding_perception/food_on_fork.py
```diff
@@ -0,0 +1,363 @@
...
+                    None,
+                    ParameterDescriptor(
+                        name="right_bottom_corner_x",
+                        type=ParameterType.PARAMETER_INTEGER,
+                        description="x-value of Right bottom corner of the rectangle",
+                        read_only=True,
+                    ),
+                ),
+                (
+                    "right_bottom_corner_y",
```
amalnanavati: Nit: "top-left", "bottom-right" etc. are more commonly used. Please swap the first two words of these names.

ada_feeding_perception/ada_feeding_perception/food_on_fork.py
```diff
@@ -0,0 +1,363 @@
...
+                    ParameterDescriptor(
+                        name="min_dist",
+                        type=ParameterType.PARAMETER_INTEGER,
+                        description="minimum depth to consider (note that 330 is approx distance "
+                                    "to the fork tine)",
+                        read_only=True,
+                    ),
+                ),
+                (
+                    "max_dist",
```
amalnanavati: Nit: Can you rename these to `min_depth` and `max_depth`? Distance is less clear as it can even refer to distance in pixel-space, whereas depth is unambiguous.

ada_feeding_perception/ada_feeding_perception/food_on_fork.py
```diff
@@ -0,0 +1,363 @@
...
+                        read_only=True,
+                    ),
+                ),
+                (
+                    "model_location",
+                    None,
+                    ParameterDescriptor(
+                        name="model_location",
+                        type=ParameterType.PARAMETER_STRING,
+                        description="model location",
```
amalnanavati, line 201: Need a description.

ada_feeding_perception/ada_feeding_perception/food_on_fork.py
```diff
@@ -0,0 +1,363 @@
...
+                        read_only=True,
+                    ),
+                ),
+                (
+                    "test",
+                    None,
+                    ParameterDescriptor(
+                        name="test",
+                        type=ParameterType.PARAMETER_BOOL,
+                        description="testing",
```
amalnanavati: Need a description.

ada_feeding_perception/ada_feeding_perception/food_on_fork.py
```diff
@@ -0,0 +1,363 @@
...
+                    ParameterDescriptor(
+                        name="max_dist",
+                        type=ParameterType.PARAMETER_INTEGER,
+                        description="maximum depth to consider (note that 330 is approx distance "
+                                    "to the fork tine)",
+                        read_only=True,
+                    ),
+                ),
+                (
+                    "model_location",
```
amalnanavati, line 196: Nit: `model_filepath` or `model_filename` would be more clear (depending on whether you are expecting a filepath or a filename)

ada_feeding_perception/ada_feeding_perception/food_on_fork.py
```diff
@@ -0,0 +1,363 @@
...
+                        description="testing",
+                        read_only=True,
+                    ),
+                ),
+            ],
+        )
+
+    def normalize_to_uint8(self, img):
+        """
+        Normalize the image to 0-255
```
amalnanavati: Add to the docstring that this is mostly used for debugging, to allow an image that is not uint8 (e.g., a depth image, an image of probabilities) to be visualized

ada_feeding_perception/ada_feeding_perception/food_on_fork_categorical_naive_bayes_training.py
```diff
@@ -0,0 +1,288 @@
...
+
+    for labels_path in test_folder:
+        test_labels_folder[labels_path] = os.listdir(
+            os.path.join(test_folder_path, labels_path)
+        )
+
+    return train_labels_folder, test_labels_folder, train_folder_path, test_folder_path
+
+
+def normalize_to_uint8(img):
```
amalnanavati: If you use the same function in 2+ different files, you should instead define it in `helpers.py` and import it from both files.

ada_feeding_perception/ada_feeding_perception/food_on_fork.py
```diff
@@ -0,0 +1,363 @@
...
+        Parameters:
+        ----------
+        depth_img_msg: depth image message from the subscriber
+        """
+        depth_img = None
+        try:
+            depth_img = self.bridge.imgmsg_to_cv2(depth_img_msg, "passthrough")
+        except CvBridgeError as e:
+            print(e)
+
```
amalnanavati: Return within the `except`, to avoid the need for the if statement below

ada_feeding_perception/ada_feeding_perception/food_on_fork.py
```diff
@@ -0,0 +1,363 @@
...
+        if depth_img is not None:
+            # VVVV---BELOW is the code when using single feature of only the NUMBER OF PIXELS---VVVV
```
amalnanavati: I like that you made your code general enough to use with both the logistic regression and CategoricalNB models. However, requiring someone to comment out lines of code to toggle between them is quite confusing. Instead, pass in a parameter that toggles this. For example, the parameter can be called `features` and can take two values for now: `num_pixels` and `all_pixels` (if users pass in another value, raise an exception). And then over here, have an if statement toggle between the two.

ada_feeding_perception/ada_feeding_perception/food_on_fork.py
```diff
@@ -0,0 +1,363 @@
...
+            # num_pixels = self.food_on_fork_num_pixels(depth_img)
+            # if self.test:
+            #     self.get_logger().info("num pixels: " + str(num_pixels))
+            # prediction = self.predict_food_on_fork(num_pixels)
+            # if self.test:
+            #     self.get_logger().info("Prediction Probability: " + str(prediction))
+            # ^^^^---ABOVE is the code when using single feature of only the NUMBER OF PIXELS---^^^^
+
+            # VVVV---BELOW is the code when using EACH PIXEL as a feature---VVVV
+            depth_img_copy = np.copy(depth_img)
```
amalnanavati: Why do you need to copy it? This seems like an unnecessary computation to me.

ada_feeding_perception/ada_feeding_perception/food_on_fork.py
```diff
@@ -0,0 +1,363 @@
...
+                np.logical_or(cropped_img < self.min_dist, cropped_img > self.max_dist),
+                0,
+                1,
+            ).astype("uint8")
+            cropped_img_np = np.array(img_converted)
+
+            # self.get_logger().info(str(cropped_img_np))
+            # self.get_logger().info(str(np.count_nonzero(cropped_img_np == 1)))
+            # cv.imshow("cropped", cropped_img_np * 255)
+            # cv.waitKey(40)
```
amalnanavati: It is fine to keep commented out debug statements, but only keep the most important that you actually think a future programmer may need to comment out. Of these 4 lines, decide which is the most important and just keep that. (Or delete them all)

ada_feeding_perception/ada_feeding_perception/food_on_fork.py
```diff
@@ -0,0 +1,363 @@
...
+            X_test = cropped_img_np.flatten()
```
amalnanavati: Rename to `x`. `X` typically refers to a dataset, so people will expect it to be resized. What you have here is a datapoint, not a dataset.

ada_feeding_perception/ada_feeding_perception/food_on_fork.py
```diff
@@ -0,0 +1,363 @@
...
+        # The pixels within the depth range will be true,
+        # resulting in mask = [[F, F, F, F], [F, T, T, F], [F, F, F, F]]
+        mask_img[
+            np.logical_not((self.min_dist < depth_img) & (depth_img < self.max_dist))
+        ] = False
+
+        return np.count_nonzero(mask_img)
+
+    def predict_food_on_fork(
+        self, num_pixels: int = None, X_test: np.ndarray = None
```
amalnanavati:
```suggestion
        self, num_pixels: Optional[int] = None, X_test: Optional[npt.NDArray] = None
```

In imports, add:
```
from typing import Optional
import numpy.typing as npt
```

ada_feeding_perception/ada_feeding_perception/food_on_fork.py
```diff
@@ -0,0 +1,363 @@
...
+        mask_img[pt1_row:pt2_row, pt1_col:pt2_col] = True
+        # The pixels within the depth range will be true,
+        # resulting in mask = [[F, F, F, F], [F, T, T, F], [F, F, F, F]]
+        mask_img[
+            np.logical_not((self.min_dist < depth_img) & (depth_img < self.max_dist))
+        ] = False
+
+        return np.count_nonzero(mask_img)
+
+    def predict_food_on_fork(
```
amalnanavati, line 291: We should do this conditioning on type of features only once. You already do it in `listener_callback_depth`. Then, this function should just take in `x: npt.NDArray` and pass that into the model and return the output. It shouldn't matter whether `x` has one feature or 10K. 

In general, you should make the assumption that the programmer specifically specified the rosparams (you can even check that in the load_params function if you want) and just pass in the datapoint with the num/type features they specified into the model. If they misspecified the rosparams, there will be an error, and they can address that.

ada_feeding_perception/ada_feeding_perception/food_on_fork.py
```diff
@@ -0,0 +1,363 @@
...
+            # self.get_logger().info(str(np.count_nonzero(cropped_img_np == 1)))
+            # cv.imshow("cropped", cropped_img_np * 255)
+            # cv.waitKey(40)
+            X_test = cropped_img_np.flatten()
+
+            # Get the prediction, which is a float value
+            prediction = self.predict_food_on_fork(X_test=X_test)
+            # self.get_logger().info(str(prediction))
+            if self.test:
+                self.get_logger().info(str(prediction))
```
amalnanavati, line 282: Why do you have some debugging statements commented out and others guarded by `self.test`? I think you should do one or the other (`self.test` is the better one to do).

ada_feeding_perception/ada_feeding_perception/food_on_fork.py
```diff
@@ -0,0 +1,363 @@
...
+            # ^^^^---ABOVE is the code when using EACH PIXEL as a feature---^^^^
+
+            float32msg = Float32()
+            float32msg.data = prediction
+            self.publisher_depth.publish(float32msg)
+
+    def food_on_fork_num_pixels(self, depth_img: np.ndarray) -> int:
```
amalnanavati: I think you should move both `food_on_fork_num_pixels` and the code above to compute the features when you are using all pixels into `helpers.py`. Even though you only use them here, if anyone is trying to generate a new training dataset in the future, they will need to use those helper functions.

I'd also recommend renaming them to `food_on_fork_featurizer_num_pixels` and `food_on_fork_featurizer_all_pixels`. And both should take in a `depth_img: npt.NDArray`. That way, it makes it clear that these are just two different options to featurize a depth image, and makes it easy for us to add more featurization options in the future.

ada_feeding_perception/ada_feeding_perception/food_on_fork_categorical_naive_bayes_training.py
```diff
@@ -0,0 +1,288 @@
+import zipfile
```
amalnanavati: Add docstrings and such to this file. Those errors should be raised by `pylint`

ada_feeding_perception/ada_feeding_perception/food_on_fork.py
```diff
@@ -0,0 +1,363 @@
+#!/usr/bin/env python3
+"""
+This file defines the FoodOnFork node class, which listens to a topic, /camera/depth/image_rect_raw
+and uses the depth image recieved there to calculate the probability of Food on Fork. It then
+launches a topic, /food_on_fork to which the probability of Food on Fork is published.
+"""
+import time
```
amalnanavati: Imports should be groups by standard imports first, then third-party libraries, then local imports. Within each, they should be ordered in alphabetical order. See other nodes in `ada_feeding_perception` as an example. These errors will be flagged by pylint.

ada_feeding_perception/ada_feeding_perception/food_on_fork_categorical_naive_bayes_training.py
amalnanavati: In the docstring, be sure to specify usage. Should the user pass the zipped file in or should they manually unzip the file first? Also, what folder structure does this file expect? Things like that are crucial to document.

ada_feeding_perception/ada_feeding_perception/food_on_fork_categorical_naive_bayes_training.py
```diff
@@ -0,0 +1,288 @@
...
+            title="Probability that a pixel is in range given food on the fork",
+        )
+        plt_show_depth_img(
+            prob_that_pixel_is_in_range_given_fof
+            - prob_that_pixel_is_in_range_given_no_fof,
+            title="P(px in range | FoF) - P(px in range | no FoF)",
+        )
+
+    # save model
+    if use_entire_dataset:
```
amalnanavati, line 333: It seems strange that whether or not to save the model is governed by the `use_entire_dataset` arg, as opposed to its own arg.

That is okay, but must be documented (e.g., in the docstrings of this file.)

ada_feeding_perception/ada_feeding_perception/food_on_fork_categorical_naive_bayes_training.py
```diff
@@ -0,0 +1,288 @@
...
+            ax.add_patch(Rectangle((i, j), 8, 4, edgecolor="r", linewidth=1, fill=None))
+    plt.imshow(img)
+
+    plt.title(title)
+    plt.colorbar()
+    if show:
+        plt.show()
+
+
+if "__main__" == __name__:
```
amalnanavati, line 126: It's not good coding practice to have all this code in one block. Typically you should separate it into different helper functions.

I'm fine with merging it in as is because this is a script for training so won't be called often. But in general you should avoid these huge blocks of code.

ada_feeding_perception/ada_feeding_perception/food_on_fork_logistic_reg_training.py
```diff
@@ -0,0 +1,50 @@
+import pandas as pd
+import numpy as np
+from sklearn.linear_model import LogisticRegression
+import joblib
+
+
+def main():
```
amalnanavati, line 16: Clearly this training function requires very different than the other one. That is why docstrings are crucial, for people to understand details like that.

ada_feeding_perception/ada_feeding_perception/food_on_fork_logistic_reg_training.py
```diff
@@ -0,0 +1,50 @@
...
+    csv_to_read_from = "../datasets/Ross_7-11-23.csv"
```
amalnanavati: Why is this called `Ross_...`? Seems like a very unclear name -- please change it to something more descriptive.

ada_feeding_perception/config/food_on_fork.yaml
```diff
@@ -0,0 +1,19 @@
...
+    # Bottom-right point's x-val of the bounding box of the rectangle NOTE: previously, was 422 (but 425 is better ratio)
+    right_bottom_corner_x: 425
+    # Bottom-right point's y-val of the bounding box of the rectangle
+    right_bottom_corner_y: 332
+    # minimum depth to consider (note that 330 is approx distance to the fork tine)
+    min_dist: 310
+    # maximum depth to consider (note that 330 is approx distance to the fork tine)
+    max_dist: 370
+    # absolute path for the model
+    model_location: "/home/atharva2/atharvak_node_ws/src/ada_feeding/ada_feeding_perception/model/categorical_naive_bayes_without_hand_8-30-23.pkl"
```
amalnanavati: This should definitely be changed from your local file system. See `face_detection` for an example. We pass in the directory of the package's share folder using the launchfile, and then the model location can be specified relative to that. That way, the code can work regardless of the system.

amalnanavati (changes requested):

ada_feeding_perception/ada_feeding_perception/food_on_fork.py
atharva-kashyap: made some minor edits to not require the copying in all cases; it was to draw the rectangle but only needed in debugging cases.

ada_feeding_perception/ada_feeding_perception/food_on_fork_categorical_naive_bayes_training.py
atharva-kashyap: I will keep this in mind for the future.

atharva-kashyap: will keep in mind for future

ada_feeding_perception/README.md
```diff
@@ -98,3 +101,60 @@ Launch the web app along with all the other nodes (real or dummy) as documented
...
+The Food on Fork node gets launched automatically when the Ada Feeding perception is launched. As such, after 
+launching the perception nodes, one can perform `ros2 topic echo /food_on_fork` to be able to listen to the 
+confidence being published onto the topic. Note that 0.0 indicates `no_food` and 1.0 indicates `food` on the fork. 
+Additionally, the web application has a toggle button that can get triggered on or off to be able to listen to the  
+`/food_on_fork` topic. More documentation for the webapp can be found on its repository. 
+
+There are a couple models that can be used to determine Food on Fork. They are as follows:
+> Note that "frustum" is a 3D space around the fork
```
amalnanavati, line 113: Not necessary, but consider linking a page explaining a frustrum, e.g., https://en.wikipedia.org/wiki/Frustum

ada_feeding_perception/ada_feeding_perception/food_on_fork.py
```diff
@@ -0,0 +1,344 @@
...
+                        read_only=True,
+                    ),
+                ),
+                (
+                    "max_depth",
+                    None,
+                    ParameterDescriptor(
+                        name="max_depth",
+                        type=ParameterType.PARAMETER_INTEGER,
+                        description="maximum depth to consider (note that 330 is approx distance "
```
amalnanavati, line 190:
```suggestion
                        description="maximum depth to consider (note that 330mm is approx distance "
```

ada_feeding_perception/ada_feeding_perception/food_on_fork.py
```diff
@@ -0,0 +1,344 @@
...
+                        read_only=True,
+                    ),
+                ),
+                (
+                    "min_depth",
+                    None,
+                    ParameterDescriptor(
+                        name="min_depth",
+                        type=ParameterType.PARAMETER_INTEGER,
+                        description="minimum depth to consider (note that 330 is approx distance "
```
amalnanavati, line 179:
```suggestion
                        description="minimum depth to consider (note that 330mm is approx distance "
```

ada_feeding_perception/ada_feeding_perception/food_on_fork.py
```diff
@@ -0,0 +1,344 @@
...
+                self.get_logger().info("num pixels: " + str(num_pixels))
+            prediction = self.predict_food_on_fork(num_pixels)
+            if self.test:
+                self.get_logger().info("Prediction Probability: " + str(prediction))
+        else:
+            # when pixels are individually considered as features
+
+            # Tuning the rectangle
+            # depth_img_copy = np.copy(depth_img)
+            # cv.rectangle(depth_img_copy, (317, 238), (445, 322), (255, 0, 0))
```
amalnanavati, line 266: If you want to keep this code in the file, plase explain what these numbers should refer to. e.g., consider making each of them variables that correspond to the parameter names, and then commenting out the whole block of code including those variable assignments

ada_feeding_perception/ada_feeding_perception/food_on_fork.py
amalnanavati: I don't think this is resolved. You still have two different parameters you pass in here. Instead, you should pass in a single parameter, `x_test`, which is an ndarray of size `(1, num_features)`. Then, this function should just consist of

```
prediction_prob = self.model.predict_proba(x_test)
return float(prediction_prob[0][1])
```

ada_feeding_perception/ada_feeding_perception/food_on_fork.py
```diff
@@ -0,0 +1,344 @@
...
+        if self.single_feature:
+            # Only number of pixels is the input (single feature):
+            num_pixels = food_on_fork_featurizer_num_pixels(
+                depth_img,
+                self.left_top_corner,
+                self.right_bottom_corner,
+                self.min_depth,
+                self.max_depth,
+            )
+            if self.test:
```
amalnanavati, line 256: Building off of my comment on `predict_food_on_fork`, you should generate the different `x_test`s in the conditional, but then logging the `x_test`, getting the prediction, and logging the prediction should all happen **outside of the conditional**. That way you can reduce the amount of code that is copied-and-pasted (currently, you basically have two versions of the same code in each conditional for logging and generating the prediction.)

ada_feeding_perception/ada_feeding_perception/food_on_fork.py
```diff
@@ -0,0 +1,344 @@
...
+            return float(prediction_prob[0][1])
+
+        return -5.0
+
+
+def main(args=None):
+    print("Running Food On Fork Node")
+    rclpy.init(args=args)
+    food_on_fork = FoodOnFork()
+    executor = MultiThreadedExecutor()
```
amalnanavati, line 336: Since you only have one subscriber, you don't need the MultiThreadedExecutor. Please change to the [SingleThreadedExecutor](https://docs.ros2.org/latest/api/rclpy/api/execution_and_callbacks.html#rclpy.executors.SingleThreadedExecutor)

ada_feeding_perception/config/food_on_fork.yaml
```diff
@@ -0,0 +1,22 @@
...
+    # minimum depth to consider (note that 330 is approx distance to the fork tine)
+    min_depth: 310
+    # maximum depth to consider (note that 330 is approx distance to the fork tine)
+    max_depth: 370
+    # relative path for the model
+    model_location: model/categorical_naive_bayes_aligned_depth_10-17-23.pkl
+    # single feature or not
+    single_feature: False
+    # testing
+    test: False
```
amalnanavati, line 22: It is convention for every file to end with a newline. Please add one here: https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

ada_feeding_perception/datasets/Crop_Images.py
```diff
@@ -0,0 +1,126 @@
...
+    cropped_img = np.copy(imageToSave)
+    left_top_x, left_top_y = rectangle_left_top
+    right_bottom_x, right_bottom_y = rectangle_right_bottom
+    cropped_img = cropped_img[left_top_y:right_bottom_y, left_top_x:right_bottom_x]
+    return cropped_img
+
+
+if __name__ == "__main__":
+    rosbags_abs_location = (
+        "/media/atharvak/oldFoodManip/rosbags_aligned_depth_10-17-23/"
```
amalnanavati, line 26: 5 years from now when someone is revisiting this code, they won't have access to this path to be able to check what the folder structure is.

What would be more helpful is: (a) allowing people to pass in an argument for the paths -- don't have it hardcoded into the file; (b) describe (either in a comment in this file, or in the README) what the expected directory structure is that this file needs (e.g., a folder with subfolders that each correspond to a single rosbag). That way, even if future programmers don't have access to the same dataset, they can recreate a dataset with the same directory structure, and then your code should work.

ada_feeding_perception/datasets/ParseImagesFromRosbag.py
```diff
@@ -0,0 +1,67 @@
...
+from rosbags.rosbag2 import Reader
+from rosbags.serde import deserialize_cdr
+from cv_bridge import CvBridge, CvBridgeError
+from sensor_msgs.msg import Image
+import os
+
+
+if __name__ == "__main__":
+    rosbag_access_folder = (
+        "/media/atharvak/oldFoodManip/rosbags_aligned_depth_10-17-23/"
```
amalnanavati, line 15: Same comment here re. hardcoded paths and lack of transperancy re. the expected directory structure.

ada_feeding_perception/datasets/ParseImagesFromRosbag.py
amalnanavati, line 1: Why do you need two separate files, one to parse images and the other to crop images? why not have one script that does both?

amalnanavati (changes requested):
Mostly looks good, a few things should be changed but after you do that I envision is being ready to merge.

amalnanavati:
Closed with #169

Issue #105, opened 2023-09-26
[ROS2] Behaviors should set node in `setup` not `__init__`


amalnanavati:
Currently, all the behaviors we developed that take in a ROS node take it in as an argument to `__init__`. On the other hand, all `py_trees_ros` behaviors take the node in as a kwarg to `setup()`. This issue is to move all our behaviors to take in the node during setup, to unify it with `py_trees_ros`

egordon:
Fixed for ComputeFoodFrame in https://github.com/personalrobotics/ada_feeding/pull/114

PR #114, opened 2023-09-28
ActionServerBT Overhaul and Add Static TF Object

egordon:
# Description

As a pre-cursor to the overhaul of the MoveIt2 behaviors, this PR does a smaller overhaul of ActionServerBT (ASBT)

The key idea is that a given ASBT object should act as a generator, where `create_tree` can be called numerous times with different names and get logically distinct trees. This tree creation should be entirely independent of the action server logic so that it can be used for sub-trees.

Additionally, `send_goal`, `get_result`, and `get_feedback` should be able to be called on any result of `create_tree` in order to populate / read from the tree's data to implement the action server logic.

Individual behaviors should get their ROS2 Node from `setup` (as per #105), but the node is currently passed to the ASBT in `__init__` in order to:
1. Be used in all the AS logic (`send_goal`, `get_result`, and `get_feedback`; likely just for the node clock)
2. Be used in `create_tree` to handle only behaviors that don't yet get their ROS node in setup.

Once #105 is solved, it may make sense to remove node from init and pass it directly to the AS functions, but I don't think that's necessary.

# Testing procedure

* Test all MoveTo trees as documented in README (Note: AcquireFood should now fail with a 5s Timeout given an empty message, as there is no valid TF between `world` and empty string)
* Test `AcquireFood` as documented in #102 
* Make sure to run `AcquireFood` twice to verify that `MoveToVisitor` is re-initialized (feedback planning_time should start again at 0).

# Before Merging
- [ ] `Squash & Merge`


ada_feeding/ada_feeding/helpers.py
```diff
@@ -92,6 +93,87 @@ def quat_between_vectors(vec_from: Vector3, vec_to: Vector3) -> Quaternion:
     return ret
 
 
+def add_update_static_tf(
```
amalnanavati: Do you ever intend to let a behavior get the transforms? (e.g., to get the food frame?) If so, I think you should make an analogous getter function for this, as opposed to letting behaviors do it themselves. (To ensure locking logic and such is handled correctly)

amalnanavati: Name nit: `add_update_static_tf` seems confusing to me, why not just `set_static_tf`? And in the docstring you can document that if there is already a static tf for that key, it will be overridden (or you can make `override` a boolean flag)

ada_feeding/ada_feeding/visitors/__init__.py
```diff
@@ -0,0 +1,6 @@
+"""
+This package contains custom BT Visitors that are used in the Ada Feeding
+project. All of these extend VisitorBase.
+"""
+
+from .moveit2_visitor import MoveIt2Visitor
```
amalnanavati: I think this should be called `MoveToVisitor`, since it is specifically for all the MoveTo trees, right?

ada_feeding/ada_feeding/visitors/moveit2_visitor.py
amalnanavati, line 1: Generally looks good, but how is initialization handled for visitors?

Related to the above point, when you test this as part of AcquireFood, have at least one test that involves running AcquireFood twice, to ensure the visitor is reset properly.

ada_feeding/ada_feeding/helpers.py
egordon: Nope, behaviors should get transforms through the TransformListener (`get_tf_object()` later in the file).

egordon: Okay on the name, no need for override boolean.

ada_feeding/ada_feeding/visitors/__init__.py
egordon: I agree specifically because it is for the MoveTo-style action interface.

ada_feeding/ada_feeding/visitors/moveit2_visitor.py
egordon: Good call, I'll test this.

Added a reinit function that will be called if the tree already has a MoveToVisitor on `send_goal`.

ada_feeding/ada_feeding/helpers.py
amalnanavati: Got it. In that case, why do you need to store a dict of the transforms on the blackboard? If you just publish transforms directly to the Broadcaster as this function gets called, shouldn't the Broadcaster handle overriding old transforms? I'm not sure if the dict functionality is necessary.

egordon: We do need the dict.

The STB handles the latching (i.e. holding on to the last published messages), but `sendTransform` just forwards the transform / list of transforms to the publisher.

Actually handling adding transforms to a list is a feature [present in CPP](https://github.com/ros2/geometry2/blob/a73160df358c52d1f506fce21f15f8243e9b87f4/tf2_ros/src/static_transform_broadcaster.cpp#L52) but [not Python](https://github.com/ros2/geometry2/blob/humble/tf2_ros_py/tf2_ros/static_transform_broadcaster.py#L64)

I'll make an issue upstream

egordon: See https://github.com/ros2/geometry2/issues/631

ada_feeding/ada_feeding/trees/acquire_food_tree.py
```diff
@@ -6,96 +6,65 @@
...
         ### Define Tree Logic
 
         # Root Sequence
         root_seq = py_trees.composites.Sequence(
-            name="RootSequence",
+            name=name,
             memory=True,
             children=[
-                compute_food_frame,
+                py_trees.decorators.Timeout(
```
amalnanavati, line 54: I feel like the default duration of 5secs is probably too long to wait. What do you think about 1s?

amalnanavati (approved):
LGTM

PR #115, opened 2023-09-29
Move2Plan/Execute (MoveTo Behavior Overhaul)

egordon:
# Description

Adds the following functionality / overhauled behaviors:

* MoveIt2Plan: given a bunch of planning parameters returns a JointTrajectory
* * If all goal constraints are satisfied, return None and Success
* * If any path constraint is unsatisfied, return None and Failure (unless explicitly ignored)
* MoveIt2XConstraint: compile blackboard objects into the constraint kwargs dictionary expected by MoveIt2Plan
* MoveIt2Execute: Given a JointTrajectory, execute it on the robot
* Add the option for setting the debug log_level at the launch file level
* Add the default action_select policy to the ada_feeding main launch file
* Extend the AcquireFood tree to carry out MoveAbove in order to test the basic functionality of the new behaviors.

# Testing procedure

Follow the testing procedure in #102 with the following additional steps:
* Verify that the robot moves 5cm above the shown `food` frame with the tines perpendicular to the X (red) axis (i.e. constant action 0)
* Alternate between the 2 carrot images and see the robot go back and forth moving above the two foods

_Note: will be happy to take recommendations on how to test other components of the MoveIt2 behaviors, but this is sufficient for me now to continue working on the AcquireFood tree_

# Before Merging
- [ ] `Squash & Merge`

### DIFF
--- a/ada_feeding/ada_feeding/behaviors/acquisition/compute_action_constraints.py
+++ b/ada_feeding/ada_feeding/behaviors/acquisition/compute_action_constraints.py
@@ -1,0 +1,121 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""
+This module defines behaviors that take an AcquisitionSchema.msg object
+(or return from AcquisitionSelect.srv) and computes the outputs
+needed to send MoveIt2Plan calls.
+"""
+
+# Standard imports
+from typing import Union, Optional
+
+# Third-party imports
+from geometry_msgs.msg import Point
+import numpy as np
+from overrides import override
+import py_trees
+import ros2_numpy
+
+# Local imports
+from ada_feeding_msgs.srv import AcquisitionSelect
+from ada_feeding.helpers import BlackboardKey
+from ada_feeding.behaviors import BlackboardBehavior
+
+
+class ComputeActionConstraints(BlackboardBehavior):
+    """
+    Checks AcquisitionSelect response, implements stochastic
+    policy choice, and then decomposes into individual
+    BlackboardKey objects for MoveIt2 Behaviors.
+    """
+
+    # pylint: disable=arguments-differ
+    # We *intentionally* violate Liskov Substitution Princple
+    # in that blackboard config (inputs + outputs) are not
+    # meant to be called in a generic setting.
+
+    # pylint: disable=too-many-arguments
+    # These are effectively config definitions
+    # They require a lot of arguments.
+
+    def blackboard_inputs(
+        self,
+        action_select_response: Union[BlackboardKey, AcquisitionSelect.Response],
+        move_above_dist_m: Union[BlackboardKey, float] = 0.05,
+    ) -> None:
+        """
+        Blackboard Inputs
+
+        Parameters
+        ----------
+        action_select_response: response from AcquisitionSelect.srv
+        move_above_dist_m: how far from the food to start
+        """
+        # pylint: disable=unused-argument, duplicate-code
+        # Arguments are handled generically in base class.
+        super().blackboard_inputs(
+            **{key: value for key, value in locals().items() if key != "self"}
+        )
+
+    def blackboard_outputs(
+        self,
+        move_above_pose: Optional[BlackboardKey],  # Pose, in Food Frame
+    ) -> None:
+        """
+        Blackboard Outputs
+        By convention (to avoid collisions), avoid non-None default arguments.
+
+        Parameters
+        ----------
+        action_select_request (AcquisitionSelect.Request): request to send to AcquisitionSelect
+                                                           (copies mask input)
+        food_frame (geometry_msgs/TransformStamped): transform from world_frame to food_frame
+        """
+        # pylint: disable=unused-argument, duplicate-code
+        # Arguments are handled generically in base class.
+        super().blackboard_outputs(
+            **{key: value for key, value in locals().items() if key != "self"}
+        )
+
+    @override
+    def update(self) -> py_trees.common.Status:
+        # Docstring copied from @override
+
+        # Input Validation
+        if not self.blackboard_exists("action_select_response"):
+            return py_trees.common.Status.FAILURE
+        response = self.blackboard_get("action_select_response")
+        if response.status != "Success":
+            self.logger.error(f"Bad Action Select Response: {response.status}")
+            return py_trees.common.Status.FAILURE
+        prob = np.array(response.probabilities)
+        if (
+            len(response.actions) == 0
+            or len(response.actions) != len(response.probabilities)
+            or not np.isclose(np.sum(prob), 1.0)
+        ):
+            self.logger.error(f"Malformed action select response: {response}")
+            return py_trees.common.Status.FAILURE
+
+        # Sample Action
+        index = np.random.choice(np.arange(prob.size), p=prob)
+        action = response.actions[index]
+
+        ### Construct Constraints
+
+        # Re-scale pre-transform to move_above_dist_m
+        position = ros2_numpy.numpify(action.pre_transform.position)
+        if np.isclose(np.linalg.norm(position), 0.0):
+            self.logger.error(
+                f"Malformed action pre_transform: {action.pre_transform.position}"
+            )
+            return py_trees.common.Status.FAILURE
+        position = (
+            position
+            * self.blackboard_get("move_above_dist_m")
+            / np.linalg.norm(position)
+        )
+        action.pre_transform.position = ros2_numpy.msgify(Point, position)
+        self.blackboard_set("move_above_pose", action.pre_transform)
+
+        return py_trees.common.Status.SUCCESS
--- a/ada_feeding/ada_feeding/behaviors/acquisition/compute_action_constraints.py
+++ b/ada_feeding/ada_feeding/behaviors/acquisition/compute_action_constraints.py
@@ -83,6 +83,7 @@ class ComputeActionConstraints(BlackboardBehavior):
 
         # Input Validation
         if not self.blackboard_exists("action_select_response"):
+            self.logger.error("Missing Action Select Response")
             return py_trees.common.Status.FAILURE
         response = self.blackboard_get("action_select_response")
         if response.status != "Success":
--- a/ada_feeding/ada_feeding/behaviors/moveit2/moveit2_constraints.py
+++ b/ada_feeding/ada_feeding/behaviors/moveit2/moveit2_constraints.py
@@ -119,6 +119,8 @@ class MoveIt2JointConstraint(BlackboardBehavior):
         constraints = self.blackboard_get("constraints")
         if constraints is None:
             constraints = []
+        else:
+            constraints = constraints.copy()
         constraints.append(constraint)
         self.blackboard_set("constraints", constraints)
         return py_trees.common.Status.SUCCESS
@@ -232,6 +234,8 @@ class MoveIt2PositionConstraint(BlackboardBehavior):
         constraints = self.blackboard_get("constraints")
         if constraints is None:
             constraints = []
+        else:
+            constraints = constraints.copy()
         constraints.append(constraint)
         self.blackboard_set("constraints", constraints)
         return py_trees.common.Status.SUCCESS
@@ -351,6 +355,8 @@ class MoveIt2OrientationConstraint(BlackboardBehavior):
         constraints = self.blackboard_get("constraints")
         if constraints is None:
             constraints = []
+        else:
+            constraints = constraints.copy()
         constraints.append(constraint)
         self.blackboard_set("constraints", constraints)
         return py_trees.common.Status.SUCCESS
@@ -502,6 +508,8 @@ class MoveIt2PoseConstraint(BlackboardBehavior):
         constraints = self.blackboard_get("constraints")
         if constraints is None:
             constraints = []
+        else:
+            constraints = constraints.copy()
         constraints.append(constraint_orient)
         constraints.append(constraint_pos)
         self.blackboard_set("constraints", constraints)
--- a/ada_feeding/ada_feeding/behaviors/moveit2/moveit2_constraints.py
+++ b/ada_feeding/ada_feeding/behaviors/moveit2/moveit2_constraints.py
@@ -11,8 +11,8 @@ from typing import Any, Union, Optional, Dict, List, Tuple
 
 # Third-party imports
 from geometry_msgs.msg import (
-    TransformStamped,
-    Transform,
+    PoseStamped,
+    Pose,
     PointStamped,
     Point,
     QuaternionStamped,
@@ -22,7 +22,7 @@ from overrides import override
 import py_trees
 
 # Local imports
-from ada_feeding.behaviors.moveit2 import MoveIt2ConstraintType
+from ada_feeding.behaviors.moveit2.moveit2_plan import MoveIt2ConstraintType
 from ada_feeding.helpers import BlackboardKey
 from ada_feeding.behaviors import BlackboardBehavior
 
@@ -367,7 +367,7 @@ class MoveIt2PoseConstraint(BlackboardBehavior):
     Adds Pose Constraint to Blackboard Dictionary
     See pymoveit2:set_pose_goal() for more info.
     This is a direct combo of position + orientation
-    useful for Transform/TransformStamped ROS objects.
+    useful for Pose/PoseStamped ROS objects.
     """
 
     # pylint: disable=arguments-differ
@@ -381,7 +381,7 @@ class MoveIt2PoseConstraint(BlackboardBehavior):
 
     def blackboard_inputs(
         self,
-        transform: Union[BlackboardKey, Union[TransformStamped, Transform]],
+        pose: Union[BlackboardKey, Union[PoseStamped, Pose]],
         frame_id: Union[BlackboardKey, Optional[str]] = None,
         target_link: Union[BlackboardKey, Optional[str]] = None,
         tolerance_position: Union[BlackboardKey, float] = 0.001,
@@ -403,9 +403,9 @@ class MoveIt2PoseConstraint(BlackboardBehavior):
           - `frame_id` defaults to the base link
           - `target_link` defaults to end effector
 
-        Note: if transform is TransformStamped:
-          - `frame_id` is transform.header.frame_id (if not "")
-          - `target_link` is transform.child_frame_id (if not "")
+        Note: if pose is PoseStamped:
+          - `frame_id` is pose.header.frame_id (if not "")
+          - `target_link` is pose.child_frame_id (if not "")
 
         Parameters
         ----------
@@ -445,7 +445,7 @@ class MoveIt2PoseConstraint(BlackboardBehavior):
         # This is just checking all inputs, should be
         # easy to read.
         if (
-            not self.blackboard_exists("transform")
+            not self.blackboard_exists("pose")
             or not self.blackboard_exists("frame_id")
             or not self.blackboard_exists("target_link")
             or not self.blackboard_exists("tolerance_position")
@@ -457,24 +457,18 @@ class MoveIt2PoseConstraint(BlackboardBehavior):
         ):
             return py_trees.common.Status.FAILURE
 
-        transform = self.blackboard_get("transform")
+        pose = self.blackboard_get("pose")
         constraint_orient = (
             MoveIt2ConstraintType.ORIENTATION,
             {
-                "quat_xyzw": transform.transform.rotation
-                if isinstance(transform, TransformStamped)
-                else transform.rotation,
-                "frame_id": transform.header.frame_id
-                if (
-                    isinstance(transform, TransformStamped)
-                    and len(transform.header.frame_id) > 0
-                )
+                "quat_xyzw": pose.pose.orientation
+                if isinstance(pose, PoseStamped)
+                else pose.orientation,
+                "frame_id": pose.header.frame_id
+                if (isinstance(pose, PoseStamped) and len(pose.header.frame_id) > 0)
                 else self.blackboard_get("frame_id"),
-                "target_link": transform.child_frame_id
-                if (
-                    isinstance(transform, TransformStamped)
-                    and len(transform.child_frame_id) > 0
-                )
+                "target_link": pose.child_frame_id
+                if (isinstance(pose, PoseStamped) and len(pose.child_frame_id) > 0)
                 else self.blackboard_get("target_link"),
                 "tolerance": self.blackboard_get("tolerance_orientation"),
                 "weight": self.blackboard_get("weight_orientation"),
@@ -485,20 +479,14 @@ class MoveIt2PoseConstraint(BlackboardBehavior):
         constraint_pos = (
             MoveIt2ConstraintType.POSITION,
             {
-                "position": transform.transform.translation
-                if isinstance(transform, TransformStamped)
-                else transform.translation,
-                "frame_id": transform.header.frame_id
-                if (
-                    isinstance(transform, TransformStamped)
-                    and len(transform.header.frame_id) > 0
-                )
+                "position": pose.pose.position
+                if isinstance(pose, PoseStamped)
+                else pose.position,
+                "frame_id": pose.header.frame_id
+                if (isinstance(pose, PoseStamped) and len(pose.header.frame_id) > 0)
                 else self.blackboard_get("frame_id"),
-                "target_link": transform.child_frame_id
-                if (
-                    isinstance(transform, TransformStamped)
-                    and len(transform.child_frame_id) > 0
-                )
+                "target_link": pose.child_frame_id
+                if (isinstance(pose, PoseStamped) and len(pose.child_frame_id) > 0)
                 else self.blackboard_get("target_link"),
                 "tolerance": self.blackboard_get("tolerance_position"),
                 "weight": self.blackboard_get("weight_position"),
--- a/ada_feeding/ada_feeding/behaviors/moveit2/moveit2_constraints.py
+++ b/ada_feeding/ada_feeding/behaviors/moveit2/moveit2_constraints.py
@@ -1,0 +1,508 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""
+This module defines behaviors that package individual
+blackboard keys / constants into a dictionary
+of constraints that can be passed to MoveIt2Plan.
+"""
+
+# Standard imports
+from typing import Any, Union, Optional, Dict, List, Tuple
+
+# Third-party imports
+from geometry_msgs.msg import (
+    TransformStamped,
+    Transform,
+    PointStamped,
+    Point,
+    QuaternionStamped,
+    Quaternion,
+)
+from overrides import override
+import py_trees
+
+# Local imports
+from ada_feeding.behaviors.moveit2 import MoveIt2ConstraintType
+from ada_feeding.helpers import BlackboardKey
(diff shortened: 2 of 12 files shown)


ada_feeding/ada_feeding/behaviors/acquisition/compute_action_constraints.py
```diff
@@ -0,0 +1,121 @@
...
+            **{key: value for key, value in locals().items() if key != "self"}
+        )
+
+    @override
+    def update(self) -> py_trees.common.Status:
+        # Docstring copied from @override
+
+        # Input Validation
+        if not self.blackboard_exists("action_select_response"):
+            return py_trees.common.Status.FAILURE
```
amalnanavati, line 87: Nit: add a logger message here as well

ada_feeding/ada_feeding/behaviors/moveit2/moveit2_constraints.py
```diff
@@ -0,0 +1,504 @@
...
+        # Arguments are handled generically in base class.
+        super().blackboard_inputs(
+            **{key: value for key, value in locals().items() if key != "self"}
+        )
+
+    def blackboard_outputs(
+        self,
+        constraints: Optional[
+            BlackboardKey
+        ],  # List[Tuple[MoveIt2ConstraintType, Dict[str, Any]]]]
```
amalnanavati, line 77: FWIW, you can use `typing.Annotated` to include metadata in a type hint. The metadata won't be checked by a static type checker, but will be additional information available to the reader of the code. That might be a bit easier to understand than a comment (here and wherever else you use a comment in `blackboard_outputs`). https://docs.python.org/3/library/typing.html#typing.Annotated

ada_feeding/ada_feeding/behaviors/moveit2/moveit2_constraints.py
```diff
@@ -0,0 +1,504 @@
...
+        # This is just checking all inputs, should be
+        # easy to read.
+        if (
+            not self.blackboard_exists("joint_positions")
+            or not self.blackboard_exists("joint_names")
+            or not self.blackboard_exists("tolerance")
+            or not self.blackboard_exists("weight")
+            or not self.blackboard_exists("constraints")
+        ):
+            return py_trees.common.Status.FAILURE
```
amalnanavati, line 108: Nit: log error (here and in all below constraints)

ada_feeding/ada_feeding/behaviors/moveit2/moveit2_constraints.py
```diff
@@ -0,0 +1,504 @@
...
+            BlackboardKey,
+            Union[QuaternionStamped, Quaternion, Tuple[float, float, float, float]],
+        ],
+        frame_id: Union[BlackboardKey, Optional[str]] = None,
+        target_link: Union[BlackboardKey, Optional[str]] = None,
+        tolerance: Union[
+            BlackboardKey, Union[float, Tuple[float, float, float]]
+        ] = 0.001,
+        weight: Union[BlackboardKey, float] = 1.0,
+        parameterization: int = 0,  # 0: Euler, 1: Rotation Vector
```
amalnanavati, line 273: In the docstring, link to this for more details on parameterization: https://github.com/ros-planning/moveit_msgs/blob/master/msg/OrientationConstraint.msg

ada_feeding/ada_feeding/behaviors/moveit2/moveit2_plan.py
```diff
@@ -0,0 +1,641 @@
...
+        ignore_violated_path_constraints: if True, if a path constraint starts violated,
+                                          it's okay to plan without it.
+        pipeline_id: MoveIt2 Pipeline, usually "ompl"
+        planner_id: MoveIt2 planner to use
+        allowed_planning_time: Anytime planner timeout, passed to MoveIt2 Directly.
+                                NOTE: Not guaranteed, use a Timeout decorator for that.
+        max_velocity_scale: [0,1] determines max joint of the trajectory
+        max_acceleration_scale: [0,1] determines max joint acceleration of the trajectory
+        cartesian: True to use MoveIt2's cartesian planner, uses the below parameters
+        cartesian_max_step: % of the geodesic to step for each IK of the cartesian planner
```
amalnanavati: I don't think this is percent of the geodesic, I think it is literally distance in meters. https://github.com/ros-planning/moveit_msgs/blob/9756797dcbc293c7cae66541a7bc0c7acec85892/srv/GetCartesianPath.srv#L22 https://github.com/ros-planning/moveit2/blob/5dd0d5e63c9dae08976656cb1c8473ca7a34c058/moveit_core/robot_state/src/cartesian_interpolator.cpp#L116

ada_feeding/ada_feeding/behaviors/moveit2/moveit2_plan.py
```diff
@@ -0,0 +1,641 @@
...
+                        passed to the corresponding pymoveit2 function. See moveit2.py:set_X_goal()
+        path_constraints: list of constraints, each one is a pair mapping the type to the kwargs
+                        passed to the corresponding pymoveit2 function.
+                        See moveit2.py:set_path_X_constraint()
+        ignore_violated_path_constraints: if True, if a path constraint starts violated,
+                                          it's okay to plan without it.
+        pipeline_id: MoveIt2 Pipeline, usually "ompl"
+        planner_id: MoveIt2 planner to use
+        allowed_planning_time: Anytime planner timeout, passed to MoveIt2 Directly.
+                                NOTE: Not guaranteed, use a Timeout decorator for that.
```
amalnanavati: This note is particularly important, because `allowed_planning_time` is for the planner itself, but MoveGroup has additional steps it takes before and after the planner, plus there is communication time with the BT. If programmers do use a Timeout decorator, `allowed_planning_time` should be **_less than the Timeout value_**, else it will likely necessarily fail. Perhaps add a brief mention of that here?

ada_feeding/ada_feeding/behaviors/moveit2/moveit2_plan.py
```diff
@@ -0,0 +1,641 @@
...
+        max_velocity_scale: [0,1] determines max joint of the trajectory
```
amalnanavati:
```suggestion
        max_velocity_scale: [0,1] determines max joint velocity of the trajectory
```

ada_feeding/ada_feeding/behaviors/moveit2/moveit2_plan.py
```diff
@@ -0,0 +1,641 @@
...
+        max_acceleration_scale: [0,1] determines max joint acceleration of the trajectory
+        cartesian: True to use MoveIt2's cartesian planner, uses the below parameters
+        cartesian_max_step: % of the geodesic to step for each IK of the cartesian planner
+        cartesian_jump_threshold: If >0, large joint jumps between cartesian planner IKs
```
amalnanavati:
```suggestion
        cartesian_jump_threshold: If >0, large relative joint jumps between cartesian planner IKs
```

ada_feeding/ada_feeding/behaviors/moveit2/moveit2_plan.py
```diff
@@ -0,0 +1,641 @@
...
+                        See moveit2.py:set_path_X_constraint()
+        ignore_violated_path_constraints: if True, if a path constraint starts violated,
+                                          it's okay to plan without it.
+        pipeline_id: MoveIt2 Pipeline, usually "ompl"
+        planner_id: MoveIt2 planner to use
+        allowed_planning_time: Anytime planner timeout, passed to MoveIt2 Directly.
+                                NOTE: Not guaranteed, use a Timeout decorator for that.
+        max_velocity_scale: [0,1] determines max joint of the trajectory
+        max_acceleration_scale: [0,1] determines max joint acceleration of the trajectory
+        cartesian: True to use MoveIt2's cartesian planner, uses the below parameters
```
amalnanavati, line 111: No need to do this, but given how many parameters are cartesian-specific, we should consider creating a CartesianPlanConfig class for it. If you agree, maybe we can add a TODO for it in this docstring?

ada_feeding/ada_feeding/behaviors/moveit2/moveit2_plan.py
```diff
@@ -0,0 +1,641 @@
...
+        # pylint: disable=too-many-return-statements, too-many-statements, too-many-branches
+        # All raised exceptions map to a return statement.
+        # This is the main function, it is okay for it to be long.
+
+        # Lock MoveIt2 Object
+        if self.moveit2_lock.locked():
+            return py_trees.common.Status.RUNNING
+        if self.tf_lock.locked():
+            return py_trees.common.Status.RUNNING
+        with self.moveit2_lock, self.tf_lock:
```
amalnanavati, line 218: Nit: I typically prefer only blocking for the line(s) where the global resource is being used, to free it up for other threads to use.

For example, the `tf_lock` does not need to be occupied when getting the trajectory and neither the moveit2 not tf locks need to be occupied while scaling the velocity for cartesian trajectories. Freeing up these locks would allow e.g., a behavior that is listening to the TF tree in parallel to not be blocked for longer than necessary.

ada_feeding/ada_feeding/behaviors/moveit2/moveit2_plan.py
```diff
@@ -0,0 +1,641 @@
...
+                    elif constraint_type == MoveIt2ConstraintType.ORIENTATION:
+                        goals_satisfied = (
+                            goals_satisfied
+                            and self.orientation_constraint_satisfied(constraint_kwargs)
+                        )
+                        self.moveit2.set_orientation_goal(**constraint_kwargs)
+                except tf2.LookupException:
+                    # Wait on Frame Transform
+                    # Use Timeout decorator to fail
+                    return py_trees.common.Status.RUNNING
```
amalnanavati, line 296: Perhaps log a warning here with a throttle, to make it easier for programmers to debug while their code is looping?

ada_feeding/ada_feeding/behaviors/moveit2/moveit2_plan.py
```diff
@@ -0,0 +1,641 @@
...
+                except (IndexError, AttributeError):
+                    self.logger.error(f"Malformed Goal Constraint: {constraint_kwargs}")
+                    return py_trees.common.Status.FAILURE
+
+            # If all goals are satisfied, return SUCCESS, no planning necessary
```
amalnanavati, line 301: Not sure where to add it, but in some docstring it would be helpful to specify that all the goal constraints we set specify a single goal region (e.g., we take the `and` of all the goal constraints). This library doesn't currently allow multiple goal regions (e.g., the `or` of multiple constraints), even though [it is allowed in MoveIt2](https://github.com/ros-planning/moveit_msgs/blob/9756797dcbc293c7cae66541a7bc0c7acec85892/msg/MotionPlanRequest.msg#L15).

ada_feeding/ada_feeding/behaviors/moveit2/moveit2_plan.py
```diff
@@ -0,0 +1,641 @@
...
+            if goals_satisfied:
+                self.moveit2.clear_goal_constraints()
+                self.moveit2.clear_path_constraints()
```
amalnanavati: Won't this happen in `terminate` anyway, which is called as soon as this block returns success.

ada_feeding/ada_feeding/behaviors/moveit2/moveit2_plan.py
```diff
@@ -0,0 +1,641 @@
...
+                    "path_constraints"
+                ):
+                    try:
+                        if constraint_type == MoveIt2ConstraintType.JOINT:
+                            if self.joint_constraint_satisfied(constraint_kwargs):
+                                self.moveit2.set_path_joint_constraint(
+                                    **constraint_kwargs
+                                )
+                            elif not self.blackboard_get(
+                                "ignore_violated_path_constraints"
```
amalnanavati: Consider getting this a single time before the for loop, to save multiple blackboard lookup calls if it is not a local/constant.

ada_feeding/ada_feeding/behaviors/moveit2/moveit2_plan.py
amalnanavati: Same with the path constraints

ada_feeding/ada_feeding/behaviors/moveit2/moveit2_plan.py
```diff
@@ -0,0 +1,641 @@
...
+                for constraint_type, constraint_kwargs in self.blackboard_get(
+                    "path_constraints"
+                ):
+                    try:
+                        if constraint_type == MoveIt2ConstraintType.JOINT:
+                            if self.joint_constraint_satisfied(constraint_kwargs):
+                                self.moveit2.set_path_joint_constraint(
+                                    **constraint_kwargs
+                                )
+                            elif not self.blackboard_get(
```
amalnanavati: This double negative ("not ignore") took quite a while to parse, although I have now verified that it is correct. 

Perhaps a more readable/interpretable variant of this code would be:

```
elif ignore_violated_path_constraints:
    continue # act as if this constraint did not exist
else:
    paths_satisfied = False
    break
```

ada_feeding/ada_feeding/behaviors/moveit2/moveit2_plan.py
```diff
@@ -0,0 +1,641 @@
...
+                        self.logger.error(
+                            f"Malformed Goal Constraint: {constraint_kwargs}"
+                        )
+                        return py_trees.common.Status.FAILURE
+
+                # If any path constraints are unsatisfied, return FAILURE
+                if not paths_satisfied:
+                    self.moveit2.clear_goal_constraints()
+                    self.moveit2.clear_path_constraints()
+                    self.blackboard_set("trajectory", None)
```
amalnanavati: Given that `end_joint_state` is `Optional`, should we set it to None here?

ada_feeding/ada_feeding/behaviors/moveit2/moveit2_plan.py
```diff
@@ -0,0 +1,641 @@
...
+
+        # pylint: disable=protected-access
+        # We effectively need to to introspection
+        # to copy the functionality of MoveIt constraints.
+        # And this is read-only access.
+
+        curr_joint_state = self.moveit2.joint_state
+        tol = constraint_kwargs["tolerance"]
+
+        for i, des_position in enumerate(constraint_kwargs["joint_positions"]):
```
amalnanavati: I'm confused why you are only checking the `index` in `curr_joint_state` in certain cases -- there is no guarentee that the joint_names in the constraint will be ordered in the same way as the joint names in the state publisher.

I'd recommend simplifying this logic by:

1. Looping over the min len of `joint_positions` and `joint_name` (remove unnecessary conditional)
2. For every joint_name, find its corresponding index in joint_state and use that to get `curr_position`.
3. (++) Log an error and return False if `index` throws an exception.
4. (Nit) I don't think `des_position` is the most clear name, maybe `dest_position`?

ada_feeding/ada_feeding/behaviors/moveit2/moveit2_plan.py
```diff
@@ -0,0 +1,641 @@
...
+        Parameters
+        ----------
+        constraint_kwargs: See moveit2.py; dictionary with keys: "joint_positions"
+                           "joint_names", "tolerance"
+        Returns
+        -------
+        True if the constraint is satisfied
+        """
+
+        # pylint: disable=protected-access
```
amalnanavati: Unnecessary in this function?

ada_feeding/ada_feeding/behaviors/moveit2/moveit2_plan.py
```diff
@@ -0,0 +1,641 @@
...
+        Parameters
+        ----------
+        constraint_kwargs: See moveit2.py; dictionary with keys: "position"
+                           "frame_id", "target_link", "tolerance"
+        Returns
+        -------
+        True if the constraint is satisfied
+        """
+
+        # pylint: disable=protected-access
```
amalnanavati: My last commit to `pymoveit2` addressed this: https://github.com/personalrobotics/pymoveit2/blob/amaln/move_collision/pymoveit2/moveit2.py#L1717

amalnanavati: Can be removed from orientatio constraints as well

ada_feeding/ada_feeding/behaviors/moveit2/moveit2_plan.py
```diff
@@ -0,0 +1,641 @@
...
+        # We effectively need to to introspection
+        # to copy the functionality of MoveIt constraints.
+        # And this is read-only access.
+
+        tol = constraint_kwargs["tolerance"]
```
amalnanavati: Since we did `fabs` in orientation constraints, perhaps we should do it in position and joint constraints as well to provide uniform functionality?

ada_feeding/ada_feeding/behaviors/moveit2/moveit2_plan.py
```diff
@@ -0,0 +1,641 @@
...
+        action_name: The name of the action that generated this trajectory.
+        traj: The trajectory to visualize.
+        """
+
+        # pylint: disable=too-many-locals
+        # Necessary because this function saves both the image and CSV
+
+        # pylint: disable=import-outside-toplevel
+        # No need to import graphing libraries if we aren't saving the trajectory
+        import matplotlib.pyplot as plt
```
amalnanavati, line 602: Perhaps do the same for `import csv`? I don't think we use that anywhere but here

amalnanavati: And for `from ament_index_python.packages import get_package_share_directory`

ada_feeding/ada_feeding/behaviors/moveit2/moveit2_execute.py
```diff
@@ -0,0 +1,214 @@
...
+    ) -> None:
+        """
+        Blackboard Outputs
+        By convention (to avoid collisions), avoid non-None default arguments.
+
+        Parameters
+        ----------
+        error_code: specifies the manner of execution (or goal) failure
+                    See actionlib_msgs/GoalStatus and moveit_msgs/MoveItErrorCodes for more info
+                    <0: MoveItErrorCodes Failure
```
amalnanavati, line 74: Unclear to me what the numbers here refer to. Because it is a union (not an int), anyone using this will first have to check what type it is an instance of, and then once then know that they just refer to the message definition for the error codes, right? The below comments don't clarify anything for me.

ada_feeding/ada_feeding/behaviors/moveit2/moveit2_execute.py
```diff
@@ -0,0 +1,214 @@
...
+        # Docstring copied from @override
+
+        # pylint: disable=attribute-defined-outside-init, too-many-return-statements
+        # It is okay for attributes in behaviors to be
+        # defined in the setup / initialise functions.
+
+        # Handle empty trajectory
+        if (
+            not self.blackboard_exists("trajectory")
+            or self.blackboard_get("trajectory") is None
```
amalnanavati: Do we really want to handle both these cases the same? The trajectory not existing on the blackboard seems like a programmer error that should be flagged, whereas a trajectory of None is where constraints were already satisfied, so it should succee.

ada_feeding/ada_feeding/trees/acquire_food_tree.py
```diff
@@ -63,9 +72,67 @@ def create_tree(
...
+                    inputs={
+                        "action_select_response": BlackboardKey("action_response"),
+                        # Default move_above_dist_m = 0.05
+                    },
+                    outputs={
+                        "move_above_pose": BlackboardKey("move_above_pose"),
+                    },
+                ),
+                # Move Above Food
+                MoveIt2PoseConstraint(
```
amalnanavati, line 109: Even if we are switching from the decorator to behavior convention for defining constraints, we should still have a `ClearConstraints` behavior. Else, if this tree is run multiple times (e.g., to acquire multiple bites), it will keep appending the same pose constraint to the list of goal constraints.

ada_feeding/ada_feeding/trees/acquire_food_tree.py
```diff
@@ -51,9 +59,10 @@ def create_tree(
             name=name,
             memory=True,
             children=[
+                # Compute Food Frame
```
amalnanavati, line 62: We should coalesce on a style for large composites. I personally prefer defining each behavior separately and then either adding them to the sequence at the end ([as in `MoveToMouth`](https://github.com/personalrobotics/ada_feeding/blob/9f5c219391930651baa852727004ba3c6fecba20/ada_feeding/ada_feeding/trees/move_to_mouth_tree.py#L438)), or appending them to the children as you go ([as in `scoped_behavior`](https://github.com/personalrobotics/ada_feeding/blob/9f5c219391930651baa852727004ba3c6fecba20/ada_feeding/ada_feeding/idioms/scoped_behavior.py#L88), and [other idioms in `py_trees`](https://github.com/splintered-reality/py_trees/blob/devel/py_trees/idioms.py#L195)). My concern with the style you are using is too many nested lists and too much indentation.

amalnanavati (approved):
LGTM, some changes should be made but nothing that requires another review.

Re. your question on testing the other aspects of `MoveIt2Plan/Execute`, once this is merged in I'll percolate the changes to the other behaviors and trees, which should test the other aspects of it.

ada_feeding/ada_feeding/behaviors/moveit2/moveit2_execute.py
egordon: The error codes are both type int, I'll update.
It's a case where MoveItErrorCodes and GoalStatus are msg objects with constant ints defined inside. 

Basically, the return can be either, which works because when there's an error GoalStatus reports a positive number and Move it reports a negative number.

ada_feeding/ada_feeding/trees/acquire_food_tree.py
egordon: (1) Perhaps we should consider having a helper function at a higher level that clears the blackboard before each tree run

(2) I can add this for convenience, but note that the behavior takes both an input and an output constraints array, so if the input constraints are unset or None, then it will create a new one to write to the blackboard, solving the multiple runs problem

ada_feeding/ada_feeding/trees/acquire_food_tree.py
egordon: (my larger problem with this whole issue is that they're isn't a clean way to define a tree, XML does this better)

The problem with append_children is that it obfuscates the tree structure for anybody reading the code.

What I could do is define the leaf behaviors /subtrees separately, but putting them all in at the end makes it easier to browse (especially if you use an IDE that allows collapsing if code blocks, which is most of them).

ada_feeding/ada_feeding/behaviors/moveit2/moveit2_plan.py
egordon: The problem is that with statement need to be preceded by an "if locked return running" to prevent blocking the tick thread.

But that could cause update to return prematurely and lose or recalculate local state.

I think it's better to treat the entire update, which should be quick anyway, with a single with block to prevent this.

amalnanavati: Good point, I agree with the way you’ve currently written the code.

ada_feeding/ada_feeding/trees/acquire_food_tree.py
amalnanavati: (1) I don’t think we need a higher level function, because most blackboard inputs are reset in initialise. `constraints` is the notable exception, because an initialise call can’t differentiate between the case where it is getting called on a new tree, or the constraint was populated by prior behaviors.
(2) I’m not sure how this solves the multiple runs problem? I agree that if constraints is under it will initialize it. The multiple runs issue is that constraints will already be set on a second run, so it will keep appending to that.
(3) another way to address this could be to give each constraint a name/key, and make the constraint container stored on the blackboards a dict. That way it will override the constraint if it’s name was already there.

ada_feeding/ada_feeding/behaviors/moveit2/moveit2_execute.py
amalnanavati: I don’t think they are both ints. To get to the Int, users will have to call `error_code.status` if it is a GoalStatus, or `error_code.Val` if it is a MoveItErrorCode. Hence, users are already conditioning on the type of error code.

amalnanavati: EDIT: I think we are saying the same thing. I just felt like your comment made it more confusing, but I understand your perspective that pointing out in the commend that one case is always positive and the other is negative can be helpful. I think just clarifying the comment should resolve this.

ada_feeding/ada_feeding/behaviors/moveit2/moveit2_constraints.py
egordon: Right now, I don't think there's a ton of value-add

ada_feeding/ada_feeding/behaviors/moveit2/moveit2_plan.py
egordon: Yeah, looks like meters, but since it is only a single value, for rotations, steps there are 2deg/1cm.

https://github.com/ros-planning/moveit2/blob/498a9f3d659cbcf0db03cef584d4ca16468e4dab/moveit_core/robot_state/include/moveit/robot_state/cartesian_interpolator.h#L79

ada_feeding/ada_feeding/behaviors/moveit2/moveit2_plan.py
egordon: des_position == desired_position

egordon: I'm confused. I'm not making the assumption that constraint "joint_names" is the same as the publisher. that's what the call to `index` is for, and it happens every time the constraint joint_names is defined at the current index.

The problem with the proposed logic is that I don't think it handles the logic as-defined. If "joint_names"[i] exists, do the re-mapping. Otherwise, just borrow the order from the state publisher.

(In practice, for clarity, joint_names should be defined every time, but this code handles the case where it isn't).

egordon: I can simplify the logic by requiring joint_names to be either len==0 (or None) or len(joint_positions)

ada_feeding/ada_feeding/behaviors/moveit2/moveit2_execute.py
egordon: Just made sure that error_code is always the full type and not an int (I made that mistake once by returning `MoveItErrorCodes.SUCCESS`)

ada_feeding/ada_feeding/trees/acquire_food_tree.py
egordon: Discussed externally, by passing in None as a constant to the first Move2PoseConstraints, we solve the multi-run problem

ada_feeding/ada_feeding/trees/acquire_food_tree.py
egordon: No update on this right now, but the full AcquireFoodTree will likely define leaf behaviors separately.

PR #122, opened 2023-10-18
Implement Full Schema using MoveIt2Servo

egordon:
# Description

Implement the full acquisition schema.

# Testing procedure

Tested w/ Action 0 (i.e. just straight down and up).

* Pull the updated `pr_ros_controllers`: https://github.com/personalrobotics/pr_ros_controllers/pull/30
* Tested with dummy carrots in sim
* Tested with real carrots.

Observed occasional errors:
* Misses the carrot (camera calibration? or detection before robot has finished moving above plate, maybe add an 0.5s delay?)
* Delay on twist execution means that the robot does not move as far as expected during extraction. Okay for now, need to investigate later
* Carrot depth occasionally off (TODO later: verify against depth camera)

# Before Merging
- [ ] `Squash & Merge`


amalnanavati:
Re. the following:

> or detection before robot has finished moving above plate, maybe add an 0.5s delay?

The app shows the user a live video stream, so assuming the user has at least a 0.5s delay before selecting a point on the screen, that delay is essentially there.

ada_feeding/ada_feeding/trees/acquire_food_tree.py
```diff
@@ -127,8 +218,8 @@ def create_tree(
                     wait_for_server_timeout_sec=0.0,
                 ),
                 # Get MoveIt2 Constraints
-                ComputeApproachConstraints(
-                    name="ComputeApproachConstraints",
+                ComputeActionConstraints(
```
amalnanavati: This could potentially block forever (e.g., if the tf or moveit lock is not available, or if the approach frame was not registered by the TF tree). Do we want to add a Timeout for that?

ada_feeding/ada_feeding/behaviors/ros_utility.py
```diff
@@ -0,0 +1,106 @@
...
+
+        # Get Node from Kwargs
+        self.node = kwargs["node"]
+
+    @override
+    def update(self) -> py_trees.common.Status:
+        # Docstring copied from @override
+
+        # Input Validation
+        if not self.blackboard_exists("stamped_msg") or not self.blackboard_exists(
```
amalnanavati: Note that as of #124 , you can pass a list into [`blackboard_exists`](https://github.com/personalrobotics/ada_feeding/blob/ab08c2ca70973b6f64c797874a3a7eb7a5c6d159/ada_feeding/ada_feeding/behaviors/blackboard_behavior.py#L109)

ada_feeding/ada_feeding/trees/acquire_food_tree.py
```diff
@@ -9,35 +9,39 @@
...
 from ada_feeding.idioms import (
     pre_moveto_config,
     scoped_behavior,
     retry_call_ros_service,
 )
 from ada_feeding.idioms.pre_moveto_config import set_parameter_response_all_success
+from ada_feeding.trees import StartServoTree, StopServoTree
 from ada_feeding.visitors import MoveToVisitor
 from ada_feeding_msgs.action import AcquireFood
 from ada_feeding_msgs.srv import AcquisitionSelect
```
amalnanavati: Heads-up that some versions of `pylint` view imports from `ada_feeding_msgs` as "Third-party" and not as "Local." To address this, I'd recommend putting inputs from `ada_feeding_msgs` at the top of the "Local" block, to accommodate both versions of `pylint`.

ada_feeding/ada_feeding/trees/acquire_food_tree.py
```diff
@@ -82,9 +86,96 @@ def create_tree(
...
+        blackboard.register_key(key="zero_twist", access=py_trees.common.Access.WRITE)
+        blackboard.zero_twist = TwistStamped(
+            header=Header(
+                stamp=self._node.get_clock().now().to_msg(),
+                frame_id="world",
+            ),
+            twist=Twist(),
+        )
+
+        ### Define Tree Leaf Nodes
```
amalnanavati: What was the motivation for defining these leaf nodes separately, but the other leaf nodes as part of the XML-like definition?

I think having these leaf nodes be separate is a bit confusing to read since it is unclear where variables like `approach_thresh` get written to the blackboard. Whereas if you define these leaf nodes in the tree structure, then readers would already have seen that `Compute...` outputs those blackboard variables.

ada_feeding/ada_feeding/trees/acquire_food_tree.py
```diff
@@ -231,75 +325,206 @@ def create_tree(
...
+                                                                name,
+                                                                BlackboardKey("twist"),
+                                                            ]
+                                                        ),
+                                                    ),
+                                                ],  # GraspRetry children
+                                            ),  # GraspRetry Sequence
+                                        ),  # GraspRetryInfinite SuccessIsRunning
+                                    ),  # GraspServoTimeout TimeoutFromBlackboard
+                                ),  # GraspSucceed FailureIsSuccess
```
amalnanavati: This helps readability a lot -- thanks!

ada_feeding/ada_feeding/trees/acquire_food_tree.py
```diff
@@ -231,75 +325,206 @@ def create_tree(
...
+                                        name="MoveIt2ServoTimeout",
+                                        ns=name,
+                                        duration_key=BlackboardKey("duration"),
+                                        child=py_trees.decorators.SuccessIsRunning(
+                                            name="GraspRetryInfinite",
+                                            child=py_trees.composites.Sequence(
+                                                name="GraspRetry",
+                                                memory=True,
+                                                children=[
+                                                    UpdateTimestamp(
```
amalnanavati: I think we use the `UpdateTimestamp` + `publishers.FromBlackboard` sequence enough (4x in this file alone) that it makes sense to make it an idiom.

amalnanavati (approved):
LGTM. A few small comments but nothing that requires re-review.

ada_feeding/ada_feeding/trees/acquire_food_tree.py
egordon: I think I might just wrap this whole thing into a ServoMove Behavior anyway to more easily combine with MoveToVisitor, so that would encompass everything within Timeout(SuccessIsRunning(Sequence(...))) into a single value.

ada_feeding/ada_feeding/trees/acquire_food_tree.py
egordon: Incorporated into tree.

ada_feeding/ada_feeding/trees/acquire_food_tree.py
egordon: Done

egordon:
Updates tested again in sim.

PR #125, opened 2023-11-03
Implement Octomap and Misc Bug Fixes

amalnanavati:
# Description

Joint PR with [`ada_ros2`#21](https://github.com/personalrobotics/ada_ros2/pull/21).

This PR does the following:

1. Implements multiple filters on the depth image that are necessary to make Octomap work. Without it, the Octomap gets populated with noisy/hallucinated obstacles behind the fork handle.
2. Toggle off the octomap in Acquisition (Grasp + Extract) and verify that acquisition still works.
3. Add extra safeguards to StopServo. Specifically:
    1. Update the timestamp of the 0-velocity message so MoveIt Servo will actually use it (before it was stale).
    2. Add a delay after publishing the o-velocity message to ensure MoveIt Servo acts on it.
    3. Stops the servo controller, in the off-chance that MoveIt Servo did not publish the 0-velocity command to the controller. (This is to be extra-safe).
4. Other small bug/typo fixes.

# Testing procedure

**Setup**:
- Pull [`ada_ros2`#21](https://github.com/personalrobotics/ada_ros2/pull/21).
- Pull the [`amaln/attached_collision_primitive`](https://github.com/personalrobotics/pymoveit2/tree/amaln/attached_collision_primitive) branch of `pymoveit2`.
- Re-build. 
- Start the code as documented in the README (ideally on `lovelace`). 

**Testing**:
- [ ] Using the app, go through an entire bite and ensure everything works.

# Before Merging
- [ ] `Squash & Merge`


amalnanavati:
The only lingering TODO is to permanently add a spherical "bite" to the planning scene, attached to `forkTip`, so that Octomap doesn't hallucinate fixed collision voxels near the fork when it has acquired food.

ada_feeding/ada_feeding/behaviors/acquisition/compute_action_constraints.py
```diff
@@ -122,6 +122,7 @@ def setup(self, **kwargs):
     @override
     def update(self) -> py_trees.common.Status:
         # Docstring copied from @override
+        self.logger.info(f"{self.name} [{self.__class__.__name__}::update()]")
```
egordon, line 125: I would recommend moving these prints to "debug" instead of "info" if you want to put them in, but this info should already be printed by the DebugVisitor when log_level is set to debug, so perhaps you should remove.

ada_feeding/ada_feeding/behaviors/acquisition/compute_food_frame.py
```diff
@@ -139,6 +139,7 @@ def initialise(self):
     @override
     def update(self) -> py_trees.common.Status:
         # Docstring copied from @override
+        self.logger.info(f"{self.name} [{self.__class__.__name__}::update()]")
```
egordon: Same as above, DebugVisitor should handle node traversal, best not to clog up the screen with INFO messages (which will be the default for running the demo).

ada_feeding/ada_feeding/behaviors/moveit2/servo_move.py
```diff
@@ -92,6 +92,7 @@ def initialise(self) -> None:
     @override
     def update(self) -> py_trees.common.Status:
         # Docstring copied from @override
+        self.logger.info(f"{self.name} [{self.__class__.__name__}::update()]")
```
egordon: Flagging same

ada_feeding/ada_feeding/watchdog/ft_sensor_condition.py
```diff
@@ -101,7 +101,11 @@ def __ft_sensor_callback(self, msg: WrenchStamped) -> None:
                 msg.wrench.torque.z,
             ]
         )
-        ft_time = Time.from_msg(msg.header.stamp)
+        # We use the time at which the F/T message was received when checking
+        # how long has passed since receiving a message. This is to account
+        # for possible synchroneity issues or latency between the F/T node
+        # and the watchdog node.
+        ft_time = self._node.get_clock().now()
```
egordon: I'm not 100% sure this is a good idea. Is this to capture a transport delay in the FT data and then handle the case where a whole bunch of messages are queued up at once?

In that case, wouldn't the better approach be to just modify the QoS of the forque_sensor_hardware to only have one message in the queue at once? Best Effort style?

Having a stale F/T message able to feeding the watchdog could allow for dangerous force conditions to persist.

ada_feeding/config/ada_planning_scene.yaml
```diff
@@ -55,4 +60,12 @@ ada_planning_scene:
...
\ No newline at end of file
+      frame_id: root # the frame_id that the pose is relative to
+    food:
+      primitive_type: 2 # Sphere=2. See shape_msgs/SolidPrimitive.msg
+      primitive_dims: [0.05] # Sphere has 1 dim: [radius]
+      position: [0.0, 0.0, 0.0]
+      quat_xyzw: [0.0, 0.0, 0.0, 1.0]
+      attached: True
+      frame_id: forkTip # the frame_id that the pose is relative to
+      touch_links: [forkTine] # the links, in addition to frame_id, that should be ignored for collision checking
```
egordon: Nit: newline

ada_feeding/config/ada_planning_scene.yaml
```diff
@@ -55,4 +60,12 @@ ada_planning_scene:
       primitive_dims: [1.59, 1.5, 0.01] # Box has 3 dims: [x, y, z]
       position: [-0.05, 0.17, 1.05]
       quat_xyzw: [0.0, 0.0, 0.0, 1.0]
-      frame_id: root # the frame_id that the pose is relative to
\ No newline at end of file
+      frame_id: root # the frame_id that the pose is relative to
```
egordon, line 63: Just making sure that root is pre-tilt, in case the wheelchair is tilted.

ada_feeding/scripts/save_image.py
```diff
@@ -97,40 +98,45 @@ def save_image_callback(
...
+            else:
+                # Save the latest depth image
+                depth_image = self.bridge.imgmsg_to_cv2(self.latest_depth_image)
+                depth_image_filepath = self.filepath.value + "_depth.png"
+                self.get_logger().info(f"Saving depth image to {depth_image_filepath}")
+                cv2.imwrite(depth_image_filepath, depth_image)
 
         # Return a success response
-        response.success = True
+        response.success = len(response.message) == 0
```
egordon: This works, usually I like to put boolean expressions in parentheses.

ada_feeding_perception/ada_feeding_perception/republisher.py
```diff
@@ -10,14 +10,20 @@
 """
```
egordon, line 10: Note: just doing a quick browse of this file.

ada_feeding_perception/ada_feeding_perception/republisher.py
```diff
@@ -171,10 +380,158 @@ def callback(msg: Any):
             # self.get_logger().info(
             #     f"Received message on topic {i} {self.from_topics[i]}"
             # )
+            for post_processor in post_processors:
+                msg = post_processor(msg)
             self.pubs[i].publish(msg)
 
         return callback
 
+    def mask_post_processor(self, msg: Image) -> Image:
```
egordon: Wow, this is machinery to get this to work, kudos!

For readability / modularity, consider putting post-processors in their own module.

ada_feeding_perception/config/republisher.yaml
```diff
@@ -6,9 +6,39 @@ republisher:
...
+    # The binary mask used for "mask" post-processing. This mask will get scaled,
+    # possibly disproportionately, to the same of the image.
+    mask_relative_path: model/fork_handle_mask.png
+    # The size of the temporal window for the "temporal" post-processor.
+    temporal_window_size: 5
+    # The size of the square kernel for the "spatial" post-processor.
+    spatial_num_pixels: 10
+    # The min/max pixel value to be included in the threshold. For depth images, the units are mm.
+    threshold_min: 300
+    threshold_max: 2500
```
egordon: Nit: newline

egordon (approved):
Only major question is about the update F/T watchdog.

ada_feeding/ada_feeding/behaviors/acquisition/compute_action_constraints.py
amalnanavati: Ah, I fully missed the Debug Visitor and the new launchfile arg for log-level. Thanks for pointing it out!

ada_feeding/ada_feeding/watchdog/ft_sensor_condition.py
amalnanavati: Good point, I reverted it.

The reason I had made the change is I noticed one time, when the F/T sensor had been on for a while, when the messages were being received continuously but with an older timestamp. But you're right that we should understand the underlying cause for it as opposed to putting this hack-y patch.

ada_feeding/config/ada_planning_scene.yaml
amalnanavati: Yup, `root` is pre-tilt, `root_tilt` is post-tilt

PR #133, opened 2023-11-15
Compute food frame should get center of mass of mask

amalnanavati:
# Description

Earlier, the compute food frame function used the middle of the food bounding box to determine the food frame. However, if the mask is convex, that may not be accurate. Thus, this PR changes it to use the center of mass of the food mask as the food frame.

# Testing procedure

Tested it as part of the Industry Affiliates Research Day Demo and it worked (better than the earlier approach).

# Before Merging
- [ ] `Squash & Merge`


egordon:
@amalnanavati this is fine but I think the warn log adds clutter, I'll remove it on my next pr

amalnanavati:
Good point, I missed that. This PR was the result of fast-code changes to get the demo to work, so I didn't have time to be as thorough as ideal.

PR #33, opened 2023-07-12
Implemented MoveToConfiguration

amalnanavati:
# Description

This PR, in service of issue #30, does the following:
1. Implement a generic MoveToConfiguration behavior
2. Implement a MoveToConfigurationTree pytree that only consists of the above behavior.
3. Implements MoveAbovePlate, MoveToRestingPosition, and MoveToStowLocation as instances of the above pytree.

## Design Decisions
Because all robot motion actions in ada_feeding have the same return types for results and feedback, we separate the shared logic into its own parent classes:
- **Behavior**: `MoveTo` contains the code to create a MoveIt interface, initialize feedback on the blackboard, track the progress of robot planning and motion, and handle termination. The only thing it does not do is initiate a planning call and configure planning specific parameters, which is what subclasses (e.g., `MoveToConfiguration`) must do.
- **Tree**: `MoveToTree` contains code to initialize variables for feedback on the blackboard, send the goal, get feedback, and get the result. The only thing is does not do is create the actual tree to handle robot motion, which is what subclasses (e.g., `MoveToConfigurationTree`) must do.

# Testing procedure

On the real robot (using lovelace), I tested that the actions executed and canceled, both when I used the ROS command line interface and when I used the web interface.

**Initialization**:
1. Launch MoveIt:
    1. Real Robot: `ros2 launch ada_moveit demo.launch.py`
    2. Sim: `ros2 launch ada_moveit demo.launch.py sim:=mock`
2. Launch the F/T sensor:
    1. `ros2 run forque_sensor_hardware forque_sensor_hardware --ros-args -p host:=xxx.xxx.x.xx` (replace the x's with the correct IP)
3. Launch the feeding nodes: `ros2 launch ada_feeding ada_feeding_launch.xml`
4. Launch the web interface:
    1. See here for instructions to [launch the web interface](https://github.com/personalrobotics/feeding_web_interface/tree/main/feedingwebapp#usage-web-app). Before launching, change the hostname in `.env` to the IP address of the computer you are running ROS on. 
    2. Connect your phone to the same network as the device running the ROS nodes and web app.
    3. On your phone, navigate to `xxx.xxx.x.x:3000` where the x's are replaced with the IP address. You should see a green square in the top-right corner to indicate that the app is connected with the robot.

**Testing**:
1. **Command Line Interface**: For each of the following commands, run it and verify that it executes as expected. Then, run it and terminate it, and verify that the robot immediately stops moving.
[...]

PR #30, opened 2023-06-21
pytree move actions

Raidakarim:
# Description

[TODO: describe, in-detail, what issue this PR addresses and how it addresses it. Link to relevant Github Issues.]

# Testing procedure

[TODO: describe, in-detail, how you tested this. The procedure must be detailed enough for the reviewer(s) to recreate it.]

# Before opening a pull request
- [ ] Format your code using [black formatter](https://black.readthedocs.io/en/stable/) `python3 -m black .`
- [ ] Run your code through [pylint](https://pylint.readthedocs.io/en/latest/) and address (most) warnings/errors: `pylint --recursive=y .`

# Before Merging
- [ ] `Squash & Merge`

PR #135, opened 2023-11-18
Implement `servo_until` and `servo_until_pose`

amalnanavati:
# Description

Linked with [`ada_ros2`#33](https://github.com/personalrobotics/ada_ros2/pull/33) and [PRL's fork of moveit2 (branch: `amaln/main`)](https://github.com/personalrobotics/moveit2/tree/amaln/main).

Currently, we can only servo for a duration. This PR extends that by implementing an idiom, `servo_until`` allowing us to servo until an arbitrary condition is met. It further implements a version of that idiom, `servo_until_pose`, that servos until the end effector is at a target pose.

# Testing procedure

Pull the linked repositories (see above) and rebuild. Set up the code as in the README. Pull [`ada_ros2`#30](https://github.com/personalrobotics/ada_ros2/pull/30).

- [x] Sim:
    - [x] `ros2 action send_goal /MoveToStagingConfiguration ada_feeding_msgs/action/MoveTo "{}" --feedback`
    - [x] `ros2 action send_goal /MoveToMouth ada_feeding_msgs/action/MoveToMouth "{}" --feedback`. Verify that it smoothly servos to 2.5cm in front of the mouth.
- [x] Real:
    - [x] `ros2 action send_goal /MoveToStagingConfiguration ada_feeding_msgs/action/MoveTo "{}" --feedback`
    - [x] `ros2 action send_goal /MoveToMouth ada_feeding_msgs/action/MoveToMouth "{}" --feedback`. Verify that it smoothly servos to 2.5cm in front of the mouth.
    - [x] `ros2 action send_goal /MoveFromMouthToStagingConfiguration ada_feeding_msgs/action/MoveTo "{}" --feedback`

# Before Merging
- [ ] `Squash & Merge`


ada_feeding/ada_feeding/behaviors/moveit2/servo_move.py
```diff
@@ -42,17 +42,20 @@ def blackboard_inputs(
...
         Parameters
         ----------
         twist: Twist or TwistStamped to publish, with updated time stamps.
-        duration: How long to publish the twist
+        duration: How long to publish the twist. If negative, then run forever.
         pub_topic: Where to publish servo TwistStamped messages
         pub_qos: QoS for publisher
         default_frame_id: frame_id to use if Twist type is provided.
+        curr_distance: If None, use time as distance. Otherwise, use this
+            BlackboardKey to store the remaining distance.
```
amalnanavati: nit: "this BlackboardKey stores..." (not "use" since this node doesn't write to it.)

ada_feeding/ada_feeding/trees/move_to_mouth_tree.py
```diff
@@ -82,12 +86,12 @@ def __init__(
...
         wheelchair_collision_object_id: str = "wheelchair_collision",
-        force_threshold: float = 4.0,
-        torque_threshold: float = 4.0,
+        force_threshold: float = 1.0,
+        torque_threshold: float = 1.0,
         allowed_face_distance: Tuple[float, float] = (0.4, 1.25),
         face_detection_msg_timeout: float = 5.0,
         face_detection_timeout: float = 2.5,
-        plan_distance_from_mouth: float = 0.05,
+        plan_distance_from_mouth: float = 0.025,
```
amalnanavati: Self-reminder to merge in the demo-specific changes on `lovelace`

amalnanavati:
@egordon We should see how frequently a strict-servoing approach fails (e.g., due to singularities or closeness to collisions). If it fails somewhat frequently, we may still need the cartesian interpolator as a fallback. (Because unfortuantely the aforementioned failures of servo basically brings servo to a hard-stop, and nothing can proceed.)

amalnanavati:
@egordon this has been tested in real and works well ([`ada_ros2`#30](https://github.com/personalrobotics/ada_ros2/pull/30)). It provides a lot more flexibility than the cartesian interpolator (e.g., to rotate the fork at the mouth, to pass in a custom speed profile, etc.) at the cost of a small amount more jerkiness. Three discussion points:

1. Currently, on timeout this succeeds, because we use the `duration` field of `ServoMove`. Should it instead fail (e.g., using the `Timeout` decorator)? The potential issue with doing so is that sometimes it gets quite close to the target, but can't reach it since it can't execute small velocities, and therefore it times out. But perhaps the better way of resolving that issue is increasing the timeout instead?
2. What do you think about aborting on F/T trip, as opposed to waiting until timeout? The way to do that would be to have the idiom itself subscribe to the F/T sensor and fail when the reading exceeds the threshold.
3. As far as I can tell, servo doesn't account for the octomap. It moves the fork directly in collision with the octomap. This may be a big problem for transfer, but I'm not sure because the F/T threshold is set at 1N so even if it collides with any assistive technology it won't do damage. What do you think (also @taylorkf can you weigh in on this)?

taylorkf:
> @egordon this has been tested in real and works well ([`ada_ros2`#30](https://github.com/personalrobotics/ada_ros2/pull/30)). It provides a lot more flexibility than the cartesian interpolator (e.g., to rotate the fork at the mouth, to pass in a custom speed profile, etc.) at the cost of a small amount more jerkiness. Three discussion points:
> 
> 1. Currently, on timeout this succeeds, because we use the `duration` field of `ServoMove`. Should it instead fail (e.g., using the `Timeout` decorator)? The potential issue with doing so is that sometimes it gets quite close to the target, but can't reach it since it can't execute small velocities, and therefore it times out. But perhaps the better way of resolving that issue is increasing the timeout instead?
> 2. What do you think about aborting on F/T trip, as opposed to waiting until timeout? The way to do that would be to have the idiom itself subscribe to the F/T sensor and fail when the reading exceeds the threshold.
> 3. As far as I can tell, servo doesn't account for the octomap. It moves the fork directly in collision with the octomap. This may be a big problem for transfer, but I'm not sure because the F/T threshold is set at 1N so even if it collides with any assistive technology it won't do damage. What do you think (also @taylorkf can you weigh in on this)?

Hm....I don't like the idea of just crossing our fingers and hoping the arm won't run into assistive tech. I highly doubt anything could get damaged at 1N, but it might lead to a lot of errors. I can help look into why servo isn't paying attention to the octomap and see if it's fixable.

ada_feeding/ada_feeding/behaviors/moveit2/servo_move.py
```diff
@@ -42,17 +42,20 @@ def blackboard_inputs(
...
 
         Parameters
         ----------
         twist: Twist or TwistStamped to publish, with updated time stamps.
-        duration: How long to publish the twist
+        duration: How long to publish the twist. If negative, then run forever.
         pub_topic: Where to publish servo TwistStamped messages
         pub_qos: QoS for publisher
         default_frame_id: frame_id to use if Twist type is provided.
+        curr_distance: If None, use time as distance. Otherwise, this
```
egordon: This doesn't appear to be used at all in this Behavior. I assume that it is for the MoveItVisitor to calculate time left to move.

If so, I would completely remove it from this behavior. It just makes the class more confusing.

Same with initial_distance.

See my comment in MoveTo Visitor as an idea on how to fix the logic.

ada_feeding/ada_feeding/visitors/moveto_visitor.py
```diff
@@ -172,18 +172,10 @@ def run(self, behaviour: py_trees.behaviour.Behaviour) -> None:
         ):
             # Flip to execute
             if self.feedback.is_planning:
-                self.start_time = self.node.get_clock().now()
+                self.start_time = behaviour.start_time
```
egordon: I would update this class as follows:

(1) Keep a reference to the running Execute behavior (e.g. ServoMove, MoveIt2Execute) and only execute the IsPlanning logic (i.e. under `else` below on Line 179) when that running Execute behavior returns Success/Failure. This handles sequences/selectors w/o memory so we don't have to worry about planning/execute switching.

(2) Isolate the motion initial/current distance logic in here so it doesn't pollute the tree.

Look for an ApplyTransform returning SUCCESS with the linear distance output.

If we are planning, set it to motion initial distance (it doesn't matter if we give "garbage" values in motion if we are planning, since by spec those values are meaningless if we are planning anyway). If we are executing, set it to motion current distance.

Now you can remove the caching logic from ServoMove, which isn't relevant to that behavior anyway.

ada_feeding/ada_feeding/behaviors/ros/tf.py
```diff
@@ -428,11 +349,41 @@ def update(self) -> py_trees.common.Status:
             if target_frame is not None:
                 # Compute the transform from the TF tree
                 try:
-                    transformed_msg = self.tf_buffer.transform(
-                        stamped_msg,
-                        target_frame,
-                        timeout,
-                    )
+                    if isinstance(stamped_msg, TwistStamped):
```
egordon: Does TransformationRegistration not include TwistStamped? You can check with `tf_buffer.registration.print_me()`

If not, we should implement it as part of the static execution (i.e. in the global scope) of `helpers`, see http://wiki.ros.org/tf2/Tutorials/Transforming%20your%20own%20datatypes

ada_feeding/ada_feeding/behaviors/ros/time.py
```diff
@@ -0,0 +1,166 @@
...
+# Third-party imports
+from overrides import override
+import py_trees
+
+# Local imports
+from ada_feeding.behaviors import BlackboardBehavior
+from ada_feeding.helpers import BlackboardKey
+
+
+class TrackHzInitialize(BlackboardBehavior):
```
egordon: I would rename this as `WriteTimeToBlackboard`, since that's in effect what you're doing. And remove num_ticks.

ada_feeding/ada_feeding/behaviors/ros/time.py
```diff
@@ -0,0 +1,166 @@
...
+
+        # Get Node from Kwargs
+        self.node = kwargs["node"]
+
+    @override
+    def update(self) -> py_trees.common.Status:
+        # Docstring copied from @override
+
+        # Input Validation
+        if not self.blackboard_exists(["num_ticks", "start_time"]):
```
egordon, line 95: Consider initializing num_ticks to 0 if it doesn't exist (as a first case).

ada_feeding/ada_feeding/idioms/servo_until.py
```diff
@@ -0,0 +1,353 @@
+#!/usr/bin/env python3
```
egordon, line 1: Not looking too closely, we already discussed this logic.

ada_feeding/ada_feeding/trees/move_from_mouth_tree.py
```diff
@@ -12,6 +12,7 @@
 from typing import List, Optional, Tuple
```
egordon, line 12: Not looking too closely, I trust your tests.

ada_feeding/ada_feeding/trees/move_to_mouth_tree.py
```diff
@@ -31,27 +31,31 @@
 from ada_feeding_msgs.action import MoveToMouth
```
egordon, line 31: Same as above file.

ada_feeding/config/ada_feeding_action_servers.yaml
```diff
@@ -131,15 +131,9 @@ ada_feeding_action_servers:
       action_type: ada_feeding_msgs.action.MoveToMouth # required
```
egordon, line 131: Very satisfying :relieved:

egordon (changes requested):
I think the MoveToVisitor logic needs a look.

It is best if at all possible to write the tree first, and then increase the complexity of the Visitor Class as needed, so all of that logic resides in one place.

ada_feeding/ada_feeding/behaviors/moveit2/servo_move.py
```diff
@@ -118,6 +117,26 @@ def update(self) -> py_trees.common.Status:
         twist.header.stamp = self.node.get_clock().now().to_msg()
         self.pub.publish(twist)
 
+        # Write the remaining distance
+        duration = self.blackboard_get("duration")
+        self.curr_distance = self.blackboard_get("curr_distance")
+        if self.curr_distance is None:
+            self.curr_distance = duration_to_float(duration) - duration_to_float(
```
egordon: @amalnanavati Bug

You need to check whether blackboard_get("duration") is a Duration or a float. If the latter, this throws an error.

ada_feeding/ada_feeding/behaviors/ros/time.py
amalnanavati: Offline discussion: Instead of TtrackHzInitialize, have a BlackboardClear and then TrackHz does the initialization the first subsequent time it is ticked

amalnanavati:
Lingering TODOs:

- [x] Merge `ros2-devel`
- [x] Address TrackHzInitialize comment
- [x] Make the tilt on MoveToMouth change-able by YAML
- [x] [Added TODO for this] Make the MoveToMouth linear speed lower as it approaches mouth.
- [x] Make MoveIt Servo respect Octomap (see [PRL's fork of moveit2 (branch: `amaln/main`)](https://github.com/personalrobotics/moveit2/tree/amaln/main))
- [x] Testing:
    - [x] Test the acquisition's duration feedback for grasp/extraction still works.
    - [x]  Test the tilt in a variety of head configurations, not just nominal ones.

amalnanavati:
@egordon ready for re-review. I did all the changes we discussed, except making the speed slower closer to the mouth (I added a TODO for that, since the current transfer speed seems to be a comfortable speed imo).

Tomorrow I'll do further testing with off-nominal face positions, and forking `moveit_servo` to get it to respect Octomap collisions. But neither of those are blockers to re-review.

amalnanavati:
Remaining TODO:
- [x] Have `servo_until` subscribe to `/servo_node/status` and fail if it gets a status of 3 or 4 (near or in collision). See status codes [here](https://github.com/ros-planning/moveit2/blob/3144e6eb555d6265ecd1240d9932122a8f78290a/moveit_ros/moveit_servo/include/moveit_servo/status_codes.h#L48).

amalnanavati:
Thoroughly tested bite transfer and bite acquisition with the current ServoUntilPose code and our branch of MoveIt Servo (updated to respect both the Octomap and AllowedCollisionMatrix) and it works well. Tried bite transfer with 5 head positions: centered, left, right, up, down. Note that when my head was far to the left, the arm was nearing a singularity, which necessitated changing the singularity thresholds (reflected in [`ada_ros2`#33](https://github.com/personalrobotics/ada_ros2/pull/33))

egordon (approved):
This was tested as well on the Dec 9 demo.

New MoveIt Visitor looks better, so I think this is safe to merge.

amalnanavati:
Sounds good, I just did the final change: allowing `ServoMove` to subscribe to the `/servo_node/status` topic and terminate on certain statuses (e.g., on/near collision, in singularity). I've tested that change in sim, and will test in real before merging.

amalnanavati:
Tested on the robot with the latest servo status change (e.g., verify that when i put my hand in front of the robot during bite transfer, it stops before collision). Merging in.

PR #136, opened 2023-11-21
Updates From 11/16 Demo

amalnanavati:
# Description

This PR contains most of the changes we made for the 11/16 demo, with the following _**exceptions**_:

1. We permanently removed the table collision object in `ada_planning_scene.yaml`, but that is not in this PR.
2. We moved the wheelchair_collision body down 5 cm (final `z` of `-0.10`) in `ada_planning_scene.yaml` because servo was too close to it and slowed down.

Both those changes had to do with table height inconsistencies between the planning scene and the real world. Although we can keep making those changes for user studies if need be, I'm sure there is a better way to align the planning scene with the real world than permanently removing/shifting collision objects. We should keep thinking about this.

# Testing procedure

Tested during the 11/16 demo.

- [ ] Format your code using [black formatter](https://black.readthedocs.io/en/stable/) `python3 -m black .`

# Before Merging
- [ ] `Squash & Merge`


egordon (approved):
Seems like mostly formatting. Checked w/ black, looks good.

PR #141, opened 2023-12-13
Made SegmentFromPoint return the RGB image

### LINKED ISSUE #138: [ROS2] SegmentFromPoint should return an RGB image
as an RGB image within the ROI.

amalnanavati:
# Description

In service of #138, this PR expands the `Mask` message type to return an RGB image as well, so that bite acquisition no longer has to get the latest RGB image for the featurizer.

Note, I use a `CompressedImage` type because it is easier to render than on the web app side. This change will enable the web app to render one static image per mask, as opposed to three video streams.

# Testing procedure

Invoked SegmentFromPoint from the web app, received the `rgb_image` as part of the response, rendered the image in the browser (e.g., by prepending `data:image/jpeg;base64,` to the data string) and verifying that the rendered image is correct.

# Before Merging
- [ ] `Squash & Merge`

### DIFF
--- a/ada_feeding_msgs/msg/Mask.msg
+++ b/ada_feeding_msgs/msg/Mask.msg
@@ -6,6 +6,8 @@ sensor_msgs/RegionOfInterest roi
 # The binary mask, of the same size of the bounding box, that specifies which
 # pixels in the bounding box are part of the segmented item
 sensor_msgs/CompressedImage mask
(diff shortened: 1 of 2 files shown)


egordon (approved):
LGTM, since this is purely addative, I expect no issues

*Except* that the pickled Masks in the dummy acquisition tester in `feeding_web_interface` now won't work, so those need to be updated.

PR #142, opened 2023-12-15
Scale the `wheelchair_collision` object based on detected face

amalnanavati:
# Description

Paired with [`ada_ros2`#34](https://github.com/personalrobotics/ada_ros2/pull/34).

## Issue

Currently, we make the `wheelchair_collision` object---meant to represent the user's body in the wheelchair---extremely large to encompass diverse body types. As a result of that, for people who are shorter, their actual head position overlaps with the `wheelchair_collision` object. This has a few consequences:

1. The Octomap doesn't register obstacles (e.g., assistive technology) around the user's face, because it thinks those obstacles are part of the `wheelchair_collision` object.
2. We have to allow collisions with the `wheelchair_collision` object during transfer.

## Fix

This PR scales the `wheelchair_collision` object based on the user's detected head pose, so it always ends around their neck. In other words, we are encoding the assumption that people whose head is lower in the wheelchair have shorter bodies than people whose heads are higher in the wheelchair. This solution means that, the area around users' perceived head will be empty, meaning:
1. Obstacles in that area will be perceived in the Octomap.
2. We no longer need to (dis)allow collisions with `wheelchair_collision` during bite transfer.

# Testing procedure

Pull the [branch `amaln/scale_collision_meshes` from our fork of `pymoveit2`](https://github.com/personalrobotics/pymoveit2/tree/amaln/scale_collision_meshes) and the branch from [`ada_ros2`#34](https://github.com/personalrobotics/ada_ros2/pull/34). Start the code as documented in the README.

- [x] Sim: 
    - [x] Run bite transfer (move to mouth and then back above the plate) 3 times. For each time, verify that the `wheelchair_collision` object shrinks/expands based on the detected head pose, and that it is never in collision with the base link of the robot.
- [x] Real: Try the below bite transfers, verify that it works as expected.
    - [x] Head in a "normal" position for you
    - [x] Head lower than "normal" for you
    - [x] Head higher than "normal" for you
    - [x] Head to the left of "normal" for you
    - [x] Head to the right of "normal" for you
    - [x] For multiple of the above, put an obstacle to the side of your face and verify it gets added to the Octomap.
        - [x] In particular, try this out when your head pose is initially within the `wheelchair_collision` in the planning scene (e.g., the previous planning attempt had your head higher and now it is lower). This will test whether the Octomap updates fast enough to add dynamic obstacles around your face for the first transfer, if your face is lower than the large `wheelchair_collision` object.
    - [x] For multiple of the above, put an obstacle in front of your face, verify it gets added to the Octomap, and servo stops before colliding with it.

# Before Merging
- [ ] `Squash & Merge`

### DIFF
--- a/ada_feeding/ada_feeding/behaviors/transfer/planning_scene.py
+++ b/ada_feeding/ada_feeding/behaviors/transfer/planning_scene.py
@@ -35,7 +35,7 @@ class ComputeWheelchairCollisionTransform(BlackboardBehavior):
     NOTE: Although this class is in theory rich enough to compute arbitrary
     updates to the wheelchair collision object, e.g., moving its x and y to be
     centered on the face, currently we don't translate it at all and only scale
-    z in (0.0, inf). This is because incorporating translations may bring the 
+    z in (0.0, inf). This is because incorporating translations may bring the
     wheelchair collision object into collision with the robot base, which is not desirable.
     """
 
--- a/ada_feeding/ada_feeding/behaviors/transfer/planning_scene.py
+++ b/ada_feeding/ada_feeding/behaviors/transfer/planning_scene.py
@@ -1,0 +1,175 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""
+This module defines behaviors that manipulate the planning scene during bite
+transfer.
+"""
+
+# Standard imports
+from typing import Optional, Union
+
+# Third-party imports
+from geometry_msgs.msg import (
+    Point,
+    PointStamped,
+    Pose,
+    PoseStamped,
+    Quaternion,
+    QuaternionStamped,
+)
+from overrides import override
+import py_trees
+from std_msgs.msg import Header
+
+# Local imports
+from ada_feeding.behaviors import BlackboardBehavior
+from ada_feeding.helpers import BlackboardKey
+
+
+class ComputeWheelchairCollisionTransform(BlackboardBehavior):
+    """
+    A behavior that takes in the detected head pose, the original head pose,
+    and the original wheelchair_collision pose, and computes the new
+    wheelchair_collision pose and scale.
+
+    NOTE: Although this class is in theory rich enough to compute arbitrary
+    updates to the wheelchair collision object, e.g., moving its x and y to be
+    centered on the face, currently we don't translate it at all and only scale
+    z in (0.0, 1.0]. This is because incorporating translations or scaling z > 1
+    may bring the wheelchair collision object into collision with the robot base,
+    which is not desirable.
+    """
+
+    # pylint: disable=arguments-differ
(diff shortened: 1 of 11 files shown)


amalnanavati:
Ok, I've thoroughly tested the bite transfer and acquisition of this, with the new servo and Octomap parameters in [`ada_ros2`#34](https://github.com/personalrobotics/ada_ros2/pull/34), and it works reliably while detecting obstacles around the mouth.

@taylorkf even after scaling the wheelchair collision, the Octomap updates before the robot moves, so it does perceive obstacles around the face (I held up the bag of carrots and Amazon Echo to test). Thus, I think it's fine to keep the planning scene updates in `MovetoMouth`, although I added a TODO to consider moving them into `ada_planning_scene.py`.

ada_feeding/ada_feeding/behaviors/transfer/planning_scene.py
```diff
@@ -0,0 +1,171 @@
...
+
+    # pylint: disable=arguments-differ
+    # We *intentionally* violate Liskov Substitution Princple
+    # in that blackboard config (inputs + outputs) are not
+    # meant to be called in a generic setting.
+
+    def blackboard_inputs(
+        self,
+        detected_head_pose: Union[BlackboardKey, PoseStamped],
+        # NOTE: The below parameters must match the ones in `ada_planning_scene.yaml`
```
egordon, line 50: In theory, shouldn't you be able to pull the parameters from the `ada_planning_scene` node? You can use this as a default, but I believe the tree logic should do a ROS param to blackboard write and then fill that in here.

Either fix or flag as an issue.

egordon (approved):
Fine given the breadth of testing, but I would always see if it is possible to minimize the anti-pattern of "change hard-coded values in 2 places" as much as possible.

ada_feeding/ada_feeding/behaviors/transfer/planning_scene.py
amalnanavati: Good point re. Getting parameters via service call.

Although what I think may be a better architecture is to do everything in ada_planning_scene.py. Currently, that code initialise the planning scene and then dies. But instead, we can have it subscribe to the results of face detection (which is only published when it is toggled on) and update the head and wheelchair collision objects. Thus, all planning scene logic is handled in ada_planning_scene.py.

What do you think? I believe there is already an in-file TODO for this change.

taylorkf: I agree that keeping all the planning scene logic in ada_planning_scene.py makes sense.

amalnanavati: Cool, I also added an issue for that, and will merge this in

PR #148, opened 2024-01-03
Acquisition Recover on Failure

### LINKED ISSUE #126: [ROS2] AcquireFood should account for robot motions since getting the mask
Currently, the `AcquireFood` action assumes the camera is in the same pose as it was when the mask was captured from the web app. Although that is mostly true, the one case where it might not be is if the app user refreshes the page while the robot is moving. As the page is unmounting, the app will terminate the action, and then as it is re-mounting, it will call the action *with the same parameters* as when it was originally called. That is because the React app cannot differentiate between the page mounting when the user just transitions to it, versus the page mounting when it refreshes, and the "robot motion" pages are designed to call the robot action as soon as they mount.

To address this, when [AcquireFood computes the food frame](https://github.com/personalrobotics/ada_feeding/blob/c212f8305d609d255546f797783ef4c722cb1d19/ada_feeding/ada_feeding/behaviors/acquisition/compute_food_frame.py#L188), it should first use the TF tree to get the transform from the camera to the base frame *at the time included in the action goal's header*, and then compute the food frame relative to that. 

(Note, there is a separate question of whether it is desirable to have the app re-call the robot action if the page is refreshed, and, if not, how we can architect the app to avoid that. However, the above seems like a good improvement to make to acquisition code regardless, and we can also make the change to app architecture if needed.)

egordon:
# Description

This PR does 2 things:

(1) If Grasp/Extract fails OR it finishes and the Resting Pose / Above Plate movement would fail due to a F/T trip, we execute a Recovery Procedure.

The Recovery Procedure first sets the F/T threshold to max (50N) an then does a Selector to loop through recovery attempts.

Currently, this procedure only makes 1 recovery attempt: Servo up 3cm and re-check the F/T trip (attempt 3 times)

Tested in sim with manual insertions of Failure right before Extraction (and verification of 3x retry by setting the FT test to 0.01N instead of the default 4N)

(2) Calculate the TF of the food frame using the timestamp from the header.

Solves #126 

Tested in sim by refreshing the webapp during food acquisition. Whenever the new goal wasn't rejected (which happened if the previous goal didn't finish terminating), the food frame did not move despite the robot camera frame having moved.

# Before Merging
- [x] `Squash & Merge`

### DIFF
--- a/ada_feeding/ada_feeding/behaviors/acquisition/compute_food_frame.py
+++ b/ada_feeding/ada_feeding/behaviors/acquisition/compute_food_frame.py
@@ -50,6 +50,7 @@ class ComputeFoodFrame(BlackboardBehavior):
         self,
         camera_info: Union[BlackboardKey, CameraInfo],
         mask: Union[BlackboardKey, Mask],
+        timestamp: Union[BlackboardKey, rclpy.time.Time] = rclpy.time.Time(),
         food_frame_id: Union[BlackboardKey, str] = "food",
         world_frame: Union[BlackboardKey, str] = "world",
     ) -> None:
@@ -60,6 +61,8 @@ class ComputeFoodFrame(BlackboardBehavior):
         ----------
         camera_info (geometry_msgs/CameraInfo): camera intrinsics matrix
         mask (ada_feeding_msgs/Mask): food context, see Mask.msg
+        timestamp (rclpy.time.Time): Timestamp for TF transformations
+                                    (default 0 for latest)
         food_frame_id (string): If len>0, TF frame to publish static transform
                                    (relative to world_frame)
         world_frame (string): ID of the TF frame to represent the food frame in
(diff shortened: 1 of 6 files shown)


amalnanavati (approved):
Looking through the code, this looks good. 

If you're going into lab in the coming days, I think an in-person test would be useful, if we can easily come up with a way to re-create the extraction failures we were seeing. I think if we move the food out of the way during acquisition, the fork will hit the table, creating high force. Further, if we re-add the table to the planning scene, and revert back to the main branch of moveit_servo, we should be able to recreate the error.

(I'm also okay with us merging this in, and we can just see if the error re-arises on in-person testing moving forward.)

egordon:
@amalnanavati To be clear: this doesn't currently fix those errors in the sense that

(1) The max force (50N) is the threshold both in the recovery and the extraction, so if we experience >50N force we are stuck regardless (like, the only way out in that case is to remove force gating)

(2) Currently, it still uses Servo for recovery, so it will fail in the 2nd case as well in-collision.

This PR does contain the framework for adding a cartesian planner fallback, but at this point I think that is lower pri than re-adding online learning, so we can do that later.

PR #159, opened 2024-01-20
Acquisition Tweaks

amalnanavati:
1. Change `compute_food_frame` to use the median depth around the center of the mask, instead of the median depth across the whole mask (which is what is contained in `Mask.msg`). This is because if the mask is a large blob (e.g., burrito bowl), the depth might not be consistent across the whole blob, so the median depth across the whole mask may be off (e.g., if the edges of the mask are higher than the center).
2. Re-enable Octomap for the motion to the Resting Configuration.


egordon (approved):
Looks fine at a quick glance. And I agree with the changes in concept.

PR #171, opened 2024-04-02
Table Detection

sriramk117:
# Description

## Overview

This PR implements a node for table detection. As of now, the system uses a fixed table position and orientation in the planning scene. This PR addresses the issue of tables having different heights by computing the pose of the detected table from a camera image and moving the table mesh object to the calculated pose in the planning scene. 

The table detection node depends on a few assumptions to accurately detect the table's pose: (1) a plate should be in the camera view for the table detection node to publish the table's pose, (2) the plate is circular and is the largest circle in the camera view, and (3) the area within the buffer radius of 50 pixels around the plate is mostly obstacle-free.

## ROS2 Interfaces

The node exposes a toggle service (toggle_table_detection) that turns a table detection publisher on and off. The publisher (table_detection) publishes a PoseStamped message consisting of the table center and the table’s orientation as a quaternion from the perspective of the camera frame. 

## Implementation Details

We use OpenCV’s Hough Circle detection to detect a circle around the plate, draw a larger concentric circle around the plate with a specific buffer radius, and capture points in the area between these two circles representative of the table. We then clean this data (remove noise and outliers), deproject these points to a point cloud with respect to the perspective of the camera frame, and then fit a plane to the table. From this table plane, we derive the center of the table and its orientation. 

We noticed that large obstructions in the immediate area around the plate (the area between the detected plate and the larger concentric circle drawn around the plate) such as a caregiver moving their hand across the plate, can significantly alter the calculated orientation of the table. Additionally, these obstructions can be pretty frequent. So, to counteract this we use a threshold for the angular distance between the table's orientation and the table's default orientation in the planning scene. If the latest calculated orientation is within the threshold, it is accepted and otherwise, it is rejected resulting in the table using the default orientation.

# Testing procedures

### Test by running web app with mock robot:

- Run the mock robot, start the web app, and run dummy nodes for perception: `python3 src/ada_feeding/start.py --sim mock`
- Terminate the dummy nodes in the perception screen session and launch the real perception nodes within that session:
   - Reattach to perception screen session: 'screen -r perception'
   - Terminate the dummy nodes via Ctrl+C
   - Launch the real perception nodes within the screen: `ros2 launch ada_feeding_perception ada_feeding_perception.launch.py`
- On the web app, click the “Start Feeding” button and stay on the bite selection page. 
- To visually test: on RViz, see that the updated table location is representative of the actual position and orientation of the table from the web app’s camera image.
   - Also, create a visualization for the pose that gets published by the /table_detection topic. 
   - Set the shape of the pose to “Axes” from “Arrow” 

### Test by running web app with real robot:

- Run the real robot and start the web app: `python3 src/ada_feeding/start.py --dev`
- On the web app, click the “Start Feeding” button and stay on the bite selection page. 
- To visually test: on RViz, see that the updated table location is representative of the actual position and orientation of the table from the web app’s camera image.
   - Also, create a visualization for the pose that gets published by the /table_detection topic. 
   - Set the shape of the pose to “Axes” from “Arrow” 

### Test plate detection with real or mock robot:

- Follow either of the testing procedures above to run the web app with the real or mock robot.
- Set the value of the "viz" parameter in the table_detection.yaml file to True.
- On the web app, click the “Start Feeding” button and stay on the bite selection page. 
- Then, on RViz, create a visualization for the Image that gets published by the /plate_detection_img topic
   - See that a circle is being drawn around the plate in the camera image so that the plate is being accurately detected by the table detection node. 

### Testing details:
The functionality of the table detection node was assessed using various tests by running the web app with the real robot.
To verify dynamic table pose detection, we slowly lift one side of the table and check if the orientation of the table mesh object in RViz updates accordingly. 

As stated in the implementation details, a threshold is used to prevent any obstructions in the camera image from significantly altering the detected table pose. To test this threshold, we place objects in the area between the plate and the larger concentric circle around the plate drawn from the plate detection algorithm (see details in the testing procedure above for creating a visualization for the plate detection images in RViz to refer to). Then, we test that the pose of the table mesh in RViz is still representative of the real table's pose from the camera's frame of perspective and does not change to a noticeably inaccurate pose. 

Another way to test the threshold is by placing your hand directly over the plate or around the plate such that much of the area between the plate and the larger concentric circle around the plate drawn from plate detection is covered by your hand, and once again seeing that the pose of the table mesh either changes to the default table pose or doesn't change significantly.

# Testing Results:
Here, we list the results of the table detection node using various camera images to test the accuracy of table detection.  

**Test 1**
Camera image with plate detection:
testcase2-fixed-plate
Table position and orientation in the planning scene:
testcase2-fixed

**Test 2**
Camera image with plate detection:
testcase3-fixed-plate
Table position and orientation in the planning scene:
testcase3-fixed

**Test 3**
Camera image with plate detection:
testcase4-fixed-plate
Table position and orientation in the planning scene:
testcase4-fixed

# Before Merging
- [ ] `Squash & Merge`


ada_feeding/scripts/ada_planning_scene.py
```diff
@@ -573,6 +618,53 @@ def update_face_detection(self) -> None:
...
+        orientation received in the latest table detection message.
+        """
+        # Get the latest table detection message
+        with self.latest_table_detection_lock:
+            if self.latest_table_detection is None:
+                return
+            msg = self.latest_table_detection
+            self.latest_table_detection = None
+
+        base_frame = "root"
```
amalnanavati: Given that at this stage two functions use this hardcoded value, you should store it as an instance attribute. i.e., in the `__init__` function, add `self.base_frame = "root"` and then change all references to `base_frame` in other functions to `self.base_frame`.

amalnanavati: Actually what would be even better is to make it a ROS2 parameter with a default value of `root` like you did with `table_object_id`. That gives us lots of flexibility to change it in the future.

ada_feeding/scripts/ada_planning_scene.py
```diff
@@ -573,6 +618,53 @@ def update_face_detection(self) -> None:
...
+            )
+        except TransformException as e:
+            self.get_logger().error(
+                f"Failed to transform the detected table center: {e}"
+            )
+            return
+
+        # Modify z position of table for planning scene object
+        # TODO: Need to figure out this value
+        detected_table_pose.pose.position.z -= 0.78
```
amalnanavati: A few points:
1. Any magic number like this should be stored in its own variable with a descriptive name and a comment describing it. For example, in this case you could call the variable `table_z_offset`.
2. Fundamentally, the value you have here is an offset to translate the detected position into the table's origin. Thus, you need to know the table's origin. You can determine that by loading the [STL](https://github.com/personalrobotics/ada_feeding/blob/ros2-devel/ada_feeding/assets/table.stl) into some 3D modeling software ([Blender](https://www.blender.org/) is a free application for this). IIRC, I believe the origin of the table STL is at the bottom of the table's feet, centered in the other two dimensions, Since the detected point is the surface of the table, this  offset should be the table height, i.e., distance from the bottom of the table's feet to the top of the table's surface.
3. Why do you only have z offset and not an x or y offset? x and y are also important, because if those are off, the table might be too close to the robot and collide with its base. Thus, you should make some assumptions about where the plate is relative to the table, and use those to have a hard-coded x and y offset. Currently, by implicitly having an x and y offset of 0, you are assuming the center of the camera image is centered on the table. Which may be accurate or sufficient, but at least it would be useful to make that assumption explicit by including those values.
4. As mentioned in an above comment, it would be better to make these hard-coded values parameters, so they are easier to change in the future. (You should still have a description describing what these values are used for)

ada_feeding_msgs/CMakeLists.txt
```diff
@@ -35,6 +35,7 @@ rosidl_generate_interfaces(${PROJECT_NAME}
   "srv/GetSettings.srv"
   "srv/SetSettings.srv"
   "srv/ToggleFaceDetection.srv"
+  "srv/GetPoseStamped.srv"
```
amalnanavati: Are we using this anymore? If not, can you think of a future scenario where we may want to use it? If not, I'd suggest removing it.

ada_feeding_msgs/srv/GetPoseStamped.srv
```diff
@@ -0,0 +1,5 @@
+# The interface for a service that returns the pose of the table.  
+---
+# std_msgs/Header header
+# geometry_msgs/Pose pose
+geometry_msgs/PoseStamped pose_stamped
```
amalnanavati: You should generally [have an empty line at the end of every file](https://stackoverflow.com/q/2287967).

ada_feeding_perception/ada_feeding_perception/helpers.py
```diff
@@ -321,7 +321,7 @@ def get_img_msg_type(
             return Image
     raise ValueError(
         f"No publisher found with img type for topic {final_topic}. "
-        "Publishers: {[str(endpoint) for endpoint in topic_endpoints]}"
+        f"Publishers: {[str(endpoint) for endpoint in topic_endpoints]}"
```
amalnanavati, line 364: Good catch :)

ada_feeding_perception/launch/ada_feeding_perception.launch.py
```diff
@@ -145,6 +145,21 @@ def generate_launch_description():
...
+    table_detection_remappings = [
+        ("~/table_detection", "/table_detection"),
+        ("~/toggle_table_detection", "/toggle_table_detection"),
+    ]
+    table_detection = Node(
+        package="ada_feeding_perception",
+        name="table_detection",
+        executable="table_detection",
+        remappings=realsense_remappings
+        + food_detection_remappings
```
amalnanavati: Why does it need `food_detection_remappings`?

amalnanavati: Looking at the code, I understand functionally why it is necessary. However, the variable name is confusing. Perhaps rename `food_detection_remappings` to `aligned_depth_remapping`, because fundamentally that is the only info it contains?

ada_feeding_perception/ada_feeding_perception/table_detection.py
```diff
@@ -0,0 +1,483 @@
+# Standard imports
```
amalnanavati, line 6: You need a docstring at the top of the module. I'm surprised pylint didn't catch that.

ada_feeding_perception/ada_feeding_perception/table_detection.py
```diff
@@ -0,0 +1,483 @@
...
+    get_img_msg_type,
+    ros_msg_to_cv2_image,
+)
+
+
+class TableDetectionNode(Node):
+    """
+    This node subscribes to the camera info, depth image, and RGB image topics and exposes a
+    service to toggle table detection on and off. When on, the node publishes
+    a 3D PoseStamped location of the center of the table with respect to the camera's frame of
```
amalnanavati: Nit: it's not the center of the table; it is the center of the camera image, right?

ada_feeding_perception/ada_feeding_perception/table_detection.py
```diff
@@ -0,0 +1,483 @@
...
+
+    def __init__(self):
+        """
+        Initialize the TableDetectionNode.
+        """
+
+        super().__init__("table_detection")
+
+        # Table plane parameters
+        self._hough_accum = 1.5
```
amalnanavati: There is no reason another class would need to use these, right? If so, you should use the double underscore instead of the single one: https://stackoverflow.com/a/1301369

ada_feeding_perception/ada_feeding_perception/table_detection.py
```diff
@@ -0,0 +1,483 @@
...
+    """
+
+    def __init__(self):
+        """
+        Initialize the TableDetectionNode.
+        """
+
+        super().__init__("table_detection")
+
+        # Table plane parameters
```
amalnanavati: Could you add more details here? Perhaps some guidelines on how to tune these (or how you tuned them) or a link to a tutorial with more info about them.

ada_feeding_perception/ada_feeding_perception/table_detection.py
amalnanavati: If these hough parameters are only used in one function, there is no need to have them be instance attributes; you can define them at the top of that function instead. Typically, instance attributes implies that it is used in multiple places.

ada_feeding_perception/ada_feeding_perception/table_detection.py
```diff
@@ -0,0 +1,483 @@
...
+        self._hough_param1 = 100  # Larger is more selective
+        self._hough_param2 = (
+            125  # Larger is more selective/decreases chance of false positives
+        )
+        self._hough_min = 75
+        self._hough_max = 200
+        self._table_buffer = 50  # Extra radius around plate to use for table detection
+
+        # Orientation calculation parameters
+        self.deproject_center_coord = 640 * 239 + 320 - 1
```
amalnanavati: Magic numbers. Please explain how you got them or what they represent. Give each constituent number its own descriptive variable name.

ada_feeding_perception/ada_feeding_perception/table_detection.py
```diff
@@ -0,0 +1,483 @@
...
+        self.dir_constant = 0.1
```
amalnanavati: By reading the variable names and the comment, I have no idea what this does. Which is poor style; please give it a more descriptive variable name and a comment describing what it does. Also, if it is only used in a particular part of the code, consider defining it there to improve readability (since the reader has more context about how it'll be used).

ada_feeding_perception/ada_feeding_perception/table_detection.py
```diff
@@ -0,0 +1,483 @@
...
+        """
+        Run table detection at a specified fixed rate. This function gets the latest RGB
+        image and depth image messages, fits a plane to the table, and calculates the
+        center and orientation of the table in the camera frame. Then, with the calculated
+        center and orientation, it publishes a PoseStamped message. The PoseStamped
+        messages will continue to get published at a specified fixed rate as table detection
+        continues to run.
+        """
+        # Create a fixed rate to run the loop on
+        # TODO: Replace the rate with correct fixed value or create paremeter for rate
```
amalnanavati: This TODO should be done (ideally, making it a parameter)

ada_feeding_perception/ada_feeding_perception/table_detection.py
```diff
@@ -0,0 +1,483 @@
...
+            [x_ref[2], y_ref[2], z_ref[2]],
+        ]
+
+        # Derive quaternion from rotation matrix
+        q = R.from_matrix(rot_matrix)
+        q = q.as_quat()
+
+        return center, q
+
+    def toggle_table_detection_callback(
```
amalnanavati, line 179: This is a small nit but could you put these callbacks before the `fit_table` function? It improves readability for the functions to be ordered in some intuitive way, potentially in the way they are referred to. The callbacks are referred to from `__init__` whereas the other functions are referred to from `run`

ada_feeding_perception/ada_feeding_perception/table_detection.py
```diff
@@ -0,0 +1,483 @@
...
+            rgb_msg = None
+            with self.latest_img_msg_lock:
+                rgb_msg = self.latest_img_msg
+
+            # Get the latest depth image
+            depth_img_msg = None
+            with self.latest_depth_img_msg_lock:
+                depth_img_msg = self.latest_depth_img_msg
+
+            # Get table depth from camera at the (u, v) coordinate (320, 240)
```
amalnanavati: This is incorrect, right? Doesn't `fit_table` return a depth array, not a depth at a specific point?

ada_feeding_perception/ada_feeding_perception/table_detection.py
```diff
@@ -0,0 +1,483 @@
...
+            cv2.HOUGH_GRADIENT,
+            self._hough_accum,
+            self._hough_min_dist,
+            param1=self._hough_param1,
+            param2=self._hough_param2,
+            minRadius=self._hough_min,
+            maxRadius=self._hough_max,
+        )
+        if circles is None:
+            return None, None, None
```
amalnanavati: This is really bad practice; Why does the function sometimes return three things (`None, None, None`) and sometimes return one thing (`table`)?

Also, *all function returns* need to be documented in the docstring. So add something like "returns None if..." to the docstring.

ada_feeding_perception/ada_feeding_perception/table_detection.py
```diff
@@ -0,0 +1,483 @@
...
+            with self.latest_img_msg_lock:
+                rgb_msg = self.latest_img_msg
+
+            # Get the latest depth image
+            depth_img_msg = None
+            with self.latest_depth_img_msg_lock:
+                depth_img_msg = self.latest_depth_img_msg
+
+            # Get table depth from camera at the (u, v) coordinate (320, 240)
+            table_depth = self.fit_table(rgb_msg, depth_img_msg)
```
amalnanavati: You need to check here for table_depth being None, since that is a possible return value of the function.

ada_feeding_perception/ada_feeding_perception/table_detection.py
```diff
@@ -0,0 +1,483 @@
...
+            param2=self._hough_param2,
+            minRadius=self._hough_min,
+            maxRadius=self._hough_max,
+        )
+        if circles is None:
+            return None, None, None
+        circles = np.round(circles[0, :]).astype("int")
+        plate_uv = (0, 0)
+        plate_r = 0
+        print(len(circles))
```
amalnanavati: Remove all these prints.

If you feel some of them might be useful for debugging, replace them with `self.get_logger().debug(...)` which [is inherited from the Node class](https://docs.ros2.org/foxy/api/rclpy/api/node.html#rclpy.node.Node.get_logger). But keep in mind, those logs have to be useful for *someone else* in the future who might be debugging your code, so you'd need a lot more details than just printing a number with no context.

ada_feeding_perception/ada_feeding_perception/table_detection.py
```diff
@@ -0,0 +1,483 @@
...
+        for x, y, r in circles:
```
amalnanavati: Add a comment here are above that specifies that you are finding *the largest* detected circle.

ada_feeding_perception/ada_feeding_perception/table_detection.py
```diff
@@ -0,0 +1,483 @@
...
+
+# Local imports
+from ada_feeding_perception.helpers import (
+    depth_img_to_pointcloud,
+    get_img_msg_type,
+    ros_msg_to_cv2_image,
+)
+
+
+class TableDetectionNode(Node):
```
amalnanavati, line 40: It's important to have some documentation, either in the class docstring, the module docstring, the PR, or perhaps multiple of them, that clearly documents the assumptions that this node is making. (e.g., the plate has to be circular and have color contrast from the table, the plate must be the largest circle in the camera view, the n pixels around the plate must be obstacle-free, etc.) 

Because those assumptions do make our system more restricted, so it is important to be very clear about what they are. That also makes it easier for a future developer to improve upon your code.

ada_feeding_perception/ada_feeding_perception/table_detection.py
```diff
@@ -0,0 +1,483 @@
...
+        plate_uv = (0, 0)
+        plate_r = 0
+        print(len(circles))
+        for x, y, r in circles:
+            print("Radius: " + str(r))
+            if r > plate_r:
+                plate_uv = (x, y)
+                plate_r = r
+
+        # Testing - Draw out the circles
```
amalnanavati: This is never used in your current code, right? In that case, it is a waste of computational resources. I'd suggest adding a `viz` boolean ROS2 parameter, and only if it is set to `true` do you do this. 

Also, in general you shouldn't rely on people (un)commenting code to get debugging capabilities. So when `viz` is true, you should instead publish the image to a ROS topic, which the person can then view in a tool like RVIZ. See face_detection for an example.

ada_feeding_perception/ada_feeding_perception/table_detection.py
```diff
@@ -0,0 +1,483 @@
...
+            # The following lines of code are commented to prevent interruption of the program
+            # while running. Uncomment to view images of the detected circles.
+            # cv2.imshow("detected circles", gray)
+            # cv2.waitKey(0)
+
+        # Create Mask for Depth Image
+        plate_mask = np.zeros(image_depth.shape)
+        cv2.circle(plate_mask, plate_uv, plate_r + self._table_buffer, 1.0, -1)
+        cv2.circle(plate_mask, plate_uv, plate_r, 0.0, -1)
+        depth_mask = (image_depth * (plate_mask).astype("uint16")).astype(float)
```
amalnanavati: Hmm why are you converting it to type float? Because the depth mask is in integer mm units, right? I'm not sure why the float conversion is necessary.

amalnanavati: Nit: the name `depth_mask` makes it seem like a mask for the depth image, where in reality it is the masked depth image.

I'd suggest renaming it to `depth_masked` for better readability.

ada_feeding_perception/ada_feeding_perception/table_detection.py
```diff
@@ -0,0 +1,483 @@
...
+
+        # Noise removal
+        kernel = np.ones((6, 6), np.uint8)
+        depth_mask = cv2.morphologyEx(depth_mask, cv2.MORPH_OPEN, kernel, iterations=3)
+        depth_mask = cv2.morphologyEx(depth_mask, cv2.MORPH_CLOSE, kernel, iterations=3)
+
+        # Remove Outliers
```
amalnanavati: It took me a while to understand this code, which indicates a problem with readability. A few points:

1. `depth_var` makes it seem like it is the variance of depth values, which is not correct. Please change the name.
2. Please add a longer description to the comment explaining that You consider outliers to be points that are more than two standard deviations away from the mean.
3. Please store the magic number 2.0 in its own variable with a descriptive name (e.g., `z_score_thresh`) to make it more readable.

ada_feeding_perception/ada_feeding_perception/table_detection.py
```diff
@@ -0,0 +1,483 @@
...
+        # TODO: Consider outlier removal using median + iqr for the future
+        depth_var = np.abs(depth_mask - np.mean(depth_mask[depth_mask > 0]))
+        depth_std = np.std(depth_mask[depth_mask > 0])
+        depth_mask[depth_var > 2.0 * depth_std] = 0
+
+        # Fit Plane: depth = a*u + b*v + c
```
amalnanavati: This comment makes it seem like this block of code would do the actual plane fitting, but actually the below block of code does that. Merge the below block of code into this one.

(In general, your code should be blocked and commented in a way that makes it easier to read.)

ada_feeding_perception/ada_feeding_perception/table_detection.py
```diff
@@ -0,0 +1,483 @@
...
+        u_grid, v_grid = np.meshgrid(u, v)
+        table = a * u_grid + b * v_grid + c
+        table = table.astype("uint16")
+
+        return table
+
+    def get_orientation(
+        self,
+        camera_info: CameraInfo,
+        table_depth: Image,
```
amalnanavati: Is it an Image? I believe it is an `npt.NDArray`...

ada_feeding_perception/ada_feeding_perception/table_detection.py
```diff
@@ -0,0 +1,483 @@
...
+        pointcloud = depth_img_to_pointcloud(
+            table_depth,
+            0,
+            0,
+            f_x=camera_info.k[0],
+            f_y=camera_info.k[4],
+            c_x=camera_info.k[2],
+            c_y=camera_info.k[5],
+        )
+        xy_d = np.concatenate(
```
amalnanavati: Whay is xy_d? This variable name seems very unclear to me.

ada_feeding_perception/ada_feeding_perception/table_detection.py
```diff
@@ -0,0 +1,483 @@
...
+        camera_info: The camera information.
+        table_depth: The depth image of the table.
+
+        Returns
+        ----------
+        center: The center coordinates of the table in the camera frame.
+        q: The quaternion representing the orientation of the table plane.
+        """
+        # Deproject the depth image to get the 3D points in the camera frame.
+        pointcloud = depth_img_to_pointcloud(
```
amalnanavati: Do you need the whole pointcloud? If you just need one point, only deproject one point.

ada_feeding_perception/ada_feeding_perception/table_detection.py
```diff
@@ -0,0 +1,483 @@
...
+            0,
+            f_x=camera_info.k[0],
+            f_y=camera_info.k[4],
+            c_x=camera_info.k[2],
+            c_y=camera_info.k[5],
+        )
+        xy_d = np.concatenate(
+            (pointcloud[:, 0, np.newaxis], pointcloud[:, 1, np.newaxis]), axis=1
+        )
+        depth_d = pointcloud[:, 2]
```
amalnanavati: Same with this variable name. Please revisit all variable names to improve readability. Think about if you were to read this code as a newcomer -- would it be clear to you?

ada_feeding_perception/ada_feeding_perception/table_detection.py
```diff
@@ -0,0 +1,483 @@
...
+        center = [
+            pointcloud[self.deproject_center_coord][0],
+            pointcloud[self.deproject_center_coord][1],
+            pointcloud[self.deproject_center_coord][2],
+        ]
+
+        # Fit Plane: z = a*x + b*y + c
+        coeffs = np.hstack((np.vstack(xy_d), np.ones((len(xy_d), 1)))).astype(float)
+        a, b, c = np.linalg.lstsq(coeffs, depth_d, rcond=None)[0]
```
amalnanavati: I don't understand -- why are you fitting a plane again?

ada_feeding_perception/ada_feeding_perception/table_detection.py
```diff
@@ -0,0 +1,483 @@
...
+        # Create Table Depth Image
+        u = np.linspace(0, depth_mask.shape[1], depth_mask.shape[1], False)
+        v = np.linspace(0, depth_mask.shape[0], depth_mask.shape[0], False)
+        u_grid, v_grid = np.meshgrid(u, v)
+        table = a * u_grid + b * v_grid + c
+        table = table.astype("uint16")
+
+        return table
+
+    def get_orientation(
```
amalnanavati: Nit: it appears this function gets both position and orientation, so `get_pose` would be a more descriptive name.

ada_feeding_perception/ada_feeding_perception/table_detection.py
```diff
@@ -0,0 +1,483 @@
...
+            + math.pow(x_ref[1] * z_ref[0] - x_ref[0] * z_ref[1], 2)
+        )
+        y_ref = [
+            (x_ref[2] * z_ref[1] - x_ref[1] * z_ref[2]) / denom,
+            (x_ref[0] * z_ref[2] - x_ref[2] * z_ref[0]) / denom,
+            (x_ref[1] * z_ref[0] - x_ref[0] * z_ref[1]) / denom,
+        ]
+
+        # Construct rotation matrix from direction vectors
+        # TODO: Test tf_transformations quaternion_from_matrix function
```
amalnanavati: Either do the TODO, or remove it if you don't intend to do it.

ada_feeding_perception/ada_feeding_perception/table_detection.py
```diff
@@ -0,0 +1,483 @@
...
+            pointcloud[self.deproject_center_coord][2],
+        ]
+
+        # Fit Plane: z = a*x + b*y + c
+        coeffs = np.hstack((np.vstack(xy_d), np.ones((len(xy_d), 1)))).astype(float)
+        a, b, c = np.linalg.lstsq(coeffs, depth_d, rcond=None)[0]
+
+        # Modify the z coordinate of the center given the fitted plane (i.e. make the center
+        # the origin of the table plane)
+        center[2] = a * center[0] + b * center[1] + c
```
amalnanavati: This is also unclear to me -- why are you translating the center?

ada_feeding_perception/ada_feeding_perception/table_detection.py
```diff
@@ -0,0 +1,483 @@
...
+
+        # Get the direction vectors of the table plane from the camera's frame of perspective
```
amalnanavati: I understand what you're doing because we discussed it thoroughly, but by reading the code itself I was very confused. This has to do with several issues:
1. Unclear variable names -- for example, what is `x_p`? What is `denom`?
2. Comments that don't align with the code. You write `direction vector` in the commend, but it took me a while to figure out that those correspond to `x_ref, y_ref, z_ref`.

Please revamp this part to improve readability.

amalnanavati (changes requested):
See my in-line comments. In terms of functionality this is mostly good (some things in fitting the position/orientation that seemed confusing). Most of my comments have to do with getitng the code production-ready.

amalnanavati:
Also, a few small things about your PR description:

First:

> This PR addresses the issue of tables having different heights and various orientations

Correct me if I’m wrong, but this PR does not address the issue of tables having various orientations, right? Since it just picks an arbitrary orientation relative to the camera, not a perceived table orientation.

Second, your test order is incorrect / unnecessary. Because you launch perception nodes, but then run the startup script, which also launches perception nodes. So either you’ll have two copies of the perception nodes running, or the first copy will be terminated as soon as the second copy launches.

Please edit the PR description accordingly

start.py
```diff
@@ -265,7 +283,10 @@ async def main(args: argparse.Namespace, pwd: str) -> None:
     initial_start_commands = [
         f"cd {pwd}",
         "source install/setup.bash",
+        "export ROS_DOMAIN_ID=42" if args.sim == "real" else "",
```
amalnanavati: Hmm, this goes against lines 288-289.

What if instead we change the arg to `real_domain_id` with a default of 42, and then that domain ID is only used if args.sim == "real"?

Also, in terms of how to write the conditional, I prefer the construction on 288-289 compared to this, because this code unnecessarily adds an empty string in the else case, whereas the construction on 288-289 does not.

ada_feeding_perception/config/table_detection.yaml
```diff
@@ -0,0 +1,8 @@
+# NOTE: You have to change this node name if you change the node name in the launchfile.
+table_detection:
+  ros__parameters:
+    # The rate at which to publish the pose of the detected table
+    rate_hz: 3.0
+
+    # A boolean to determine whether to visualize plate detection 
+    viz: True
```
amalnanavati: Nit: can you [add a newline to the end of this file](https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline)?

ada_feeding/config/ada_planning_scene.yaml
```diff
@@ -81,3 +84,10 @@ ada_planning_scene:
       attached: True
       frame_id: forkTip # the frame_id that the pose is relative to
       touch_links: [forkTine] # the links, in addition to frame_id, that should be ignored for collision checking
+    # The offset values for the center coordinates ([x, y, z]) of the table. These 
+    # values are used to translate the detected position into the table's origin.
+    table_detection_offsets: [-0.20, -0.25, -0.79]
+    # A threshold on the distance between the default table quaternion 
+    # and the latest table quaternion to determine whether to reject 
+    # or accept the latest table quaternion
+    quat_dist_thresh: 1.396
```
amalnanavati: JW, is there some intuitive way to interpret this? Degrees tends to be more intuitive, so if you can say something like `This roughly corresponds to a range of XXX degrees in roll, pitch, and yaw` that would be helpful. I haven't worked much with quaternion differences so am not sure of the conversion, but my hunch is that it can be found with an online search.

ada_feeding/scripts/ada_planning_scene.py
```diff
@@ -34,6 +34,7 @@
 from tf2_ros.buffer import Buffer
 from tf2_ros.static_transform_broadcaster import StaticTransformBroadcaster
 from tf2_ros.transform_listener import TransformListener
+from pyquaternion import Quaternion as Q
```
amalnanavati: Don't use this. It seems like the only place you use this is lines 694-708. See below comment for suggestions on what to do instead.

ada_feeding/scripts/ada_planning_scene.py
```diff
@@ -653,15 +685,47 @@ def update_table_detection(self) -> None:
...
+        )
+        latest_table_pyquat = Q(
+            w=detected_table_pose.pose.orientation.w,
+            x=detected_table_pose.pose.orientation.x,
+            y=detected_table_pose.pose.orientation.y,
+            z=detected_table_pose.pose.orientation.z,
+        )
+
+        # Calculate the absolute distance between the latest and default table quaternions
+        quat_dist = Q.absolute_distance(default_table_pyquat, latest_table_pyquat)
```
amalnanavati: I left a comment in the param file about how it would be nice if the threshold could have intuitive units. I looked into the documentation of this library, and this function doesn't specify what units it returns. 

Instead, I'd recommend you use [the formula from here](https://math.stackexchange.com/a/90098) to get theta (in radians). That is essentially the smallest rotation you need to get from one quaternion to the other. Then, the threshold also has intuitive units (specify in the param file that it is the threshold for the smallest rotation between the current quaternion and the default quaternion, in radians.)

amalnanavati (changes requested):
This PR looks great! Thank you for doing the work and detailing your tests -- it looks a lot better 😄 

In addition to in-line comments:
- [x] Looking at the screenshots you put in, shouldn’t the top surface of the table be aligned with the bottom surface of the Octomap? It seems like there is a height difference — perhaps your z offset is wrong, or there is some transform that is slightly off somewhere?
- [x] Could you add a few sentences about what exact tests you did with the real robot? e.g., lifting the table, putting your hand across, putting items on/off the plate, etc.? Again, this is important for repeatability and to ensure that the demonstrated capabilities of his model are documented.

The main lingering work is my comment on quaternion distance and the concern about the table being lower than the Octomap. The other comments are small, and then it should be ready to merge!

ada_feeding/config/ada_planning_scene.yaml
```diff
@@ -86,8 +86,8 @@ ada_planning_scene:
       touch_links: [forkTine] # the links, in addition to frame_id, that should be ignored for collision checking
     # The offset values for the center coordinates ([x, y, z]) of the table. These 
     # values are used to translate the detected position into the table's origin.
-    table_detection_offsets: [-0.20, -0.25, -0.79]
-    # A threshold on the distance between the default table quaternion 
+    table_detection_offsets: [-0.20, -0.25, -0.735]
+    # A threshold on the angular distance between the default table quaternion
```
amalnanavati: Nit: both for this parameter and the above one, mention units in the comment. I know the position offsets are in meters. Is the angular distance threshold in radians?

ada_feeding/scripts/ada_planning_scene.py
```diff
@@ -690,24 +690,23 @@ def update_table_detection(self) -> None:
...
+            ]
         )
-        latest_table_pyquat = Q(
-            w=detected_table_pose.pose.orientation.w,
-            x=detected_table_pose.pose.orientation.x,
-            y=detected_table_pose.pose.orientation.y,
-            z=detected_table_pose.pose.orientation.z,
+        default_table_quat = np.array([0.0, 0.0, 1.0, 0.0]) # default quaternion with 180 degrees
+                                                            # rotation about the z-axis for more
+                                                            # intuitive calculation of distance
```
amalnanavati: Nit: put this comment on the line above, so it doesn't get divided over 3 lines.

ada_feeding/scripts/ada_planning_scene.py
```diff
@@ -690,24 +690,23 @@ def update_table_detection(self) -> None:
...
+                detected_table_pose.pose.orientation.z,
+                detected_table_pose.pose.orientation.w,
+            ]
         )
-        latest_table_pyquat = Q(
-            w=detected_table_pose.pose.orientation.w,
-            x=detected_table_pose.pose.orientation.x,
-            y=detected_table_pose.pose.orientation.y,
-            z=detected_table_pose.pose.orientation.z,
+        default_table_quat = np.array([0.0, 0.0, 1.0, 0.0]) # default quaternion with 180 degrees
```
amalnanavati: Huh, this seems strange to me for two reasons:

1. Shouldn't you be computing the difference between `latest_table_quat` and `self.objects[self.table_object_id].quat_xyzw`? I don't understand why you are introducing a new orientation here. Even if you need to rotate `self.objects[self.table_object_id].quat_xyzw` by 180 degrees, you should not hardcode the angle -- because in the future if we change the `self.objects[self.table_object_id].quat_xyzw` parameter in the YAML file, we should not need to also change this hardcoded angle in the code.
2. I don't understand why the 180 degree rotation is helpful / necessary / mathematically justified. Could you please clarify?

ada_feeding/scripts/ada_planning_scene.py
```diff
@@ -690,24 +690,23 @@ def update_table_detection(self) -> None:
...
+                                                            # rotation about the z-axis for more
+                                                            # intuitive calculation of distance
+        quat_dist = np.arccos(
```
amalnanavati: Nit: Please add a comment with the URL for where you got this formula from, to make it easier for future readers of the code to check your work.

start.py
```diff
@@ -283,10 +283,9 @@ async def main(args: argparse.Namespace, pwd: str) -> None:
     initial_start_commands = [
         f"cd {pwd}",
         "source install/setup.bash",
-        "export ROS_DOMAIN_ID=42" if args.sim == "real" else "",
     ]
-    if args.domain_id is not None:
-        initial_start_commands.append(f"export ROS_DOMAIN_ID={args.domain_id}")
+    if args.sim == "real":
+        initial_start_commands.append(f"export ROS_DOMAIN_ID={args.real_domain_id}")
```
amalnanavati, line 288: Good work :)

ada_feeding_msgs/CMakeLists.txt
amalnanavati: This does not exist anymore, so compilation of `ada_Feeding_msgs` will always fail. Remove this.

amalnanavati (changes requested):
Nearly there! The main important changes are to clarify/change the 180 degree rotation around the z-axis, and to remove the extra service from `ada_Feeding_msgs/CMakeLists.txt`. The other requested changes are nits.

ada_feeding/scripts/ada_planning_scene.py
```diff
@@ -690,23 +694,50 @@ def update_table_detection(self) -> None:
         detected_table_pose.pose.position.y += self.table_detection_offsets[1]
         detected_table_pose.pose.position.z += self.table_detection_offsets[2]
 
-        # Calculate the angular distance between the latest and default table quaternions
-        latest_table_quat = np.array(
+        # Store default and latest table orientations as transforms3d quaternions
+        default_table_quat = np.array(
```
amalnanavati: Nit: this doesn't need to be re-created every time this function is called, since it is constant. I'd suggest creating it once in `__init__` and then using it here.

ada_feeding/scripts/ada_planning_scene.py
```diff
@@ -690,23 +694,50 @@ def update_table_detection(self) -> None:
         detected_table_pose.pose.position.y += self.table_detection_offsets[1]
         detected_table_pose.pose.position.z += self.table_detection_offsets[2]
 
-        # Calculate the angular distance between the latest and default table quaternions
-        latest_table_quat = np.array(
+        # Store default and latest table orientations as transforms3d quaternions
```
amalnanavati: Nit: in the comment mention that the library stores quaternions as w,x,y,z

ada_feeding/scripts/ada_planning_scene.py
```diff
@@ -690,23 +694,50 @@ def update_table_detection(self) -> None:
...
         )
-        default_table_quat = np.array([0.0, 0.0, 1.0, 0.0]) # default quaternion with 180 degrees
-                                                            # rotation about the z-axis for more
-                                                            # intuitive calculation of distance
-        quat_dist = np.arccos(
-            2 * (np.dot(default_table_quat, latest_table_quat) ** 2) - 1
+
+        # Rotate the latest quaternion by 180 degrees across the z axis
+        # and store as a separate quaternion for comparison
+        z_axis_rotation = np.array([0.0, 0.0, 0.0, 1.0])
```
amalnanavati, line 724: Haha, this confused me at first since I interpreted it as x,y,z,w, so thought this was a 0 degree rotation. Until I realized it is w,x,y,z :) That's where a comment clarifying it would help.

ada_feeding/scripts/ada_planning_scene.py
```diff
@@ -690,23 +694,50 @@ def update_table_detection(self) -> None:
...
             ]
         )
-        default_table_quat = np.array([0.0, 0.0, 1.0, 0.0]) # default quaternion with 180 degrees
-                                                            # rotation about the z-axis for more
-                                                            # intuitive calculation of distance
-        quat_dist = np.arccos(
-            2 * (np.dot(default_table_quat, latest_table_quat) ** 2) - 1
+
+        # Rotate the latest quaternion by 180 degrees across the z axis
+        # and store as a separate quaternion for comparison
```
amalnanavati: Nit: maybe mention in this comment that the reason for this is that the table is symmetric around 180 degree rotations on the x axis.

amalnanavati (approved):
This looks great! I have approved it, but there were a few lingering nits regarding comments, so if you could clarify those that would be great. Once that is done, feel free to "squash and merge" this in first, and **_then_** "squash and merge" the [corresponding web app PR](https://github.com/personalrobotics/feeding_web_interface/pull/130).

ada_feeding/scripts/ada_planning_scene.py
sriramk117: Oh ok, this makes sense. Although, just to make sure, isn't the table symmetric around 180 degree rotations across the z axis not the x axis?

PR #176, opened 2024-04-30
Adding STLs For a Hospital Table Setup

hayabolotski:
Modeled after the measurements of Jonathan's real setup on 04/11/24.
Includes:
- A model for a hospital bed with and without a headboard
- A model of a person in the approximate position of how Jonathan would sit
- A model for the hospital table the robot would be mounted to and the food would be on
- A collision husk around the person


amalnanavati (approved):
These look great. Could you just do some renames to make it clear that there is a bed group and a wheelchair group?

- `body_collision.stl` --> `body_collision_in_bed.stl` (then, later, I'll change `wheelchair_collision.stl` --> `body_collision_in_wheelchair.stl` correspondingly)
- `person_laying_down.stl` --> `person_laying_in_bed.stl`

After those changes, feel free to merge it in!

PR #203, opened 2024-11-09
Semantic Labeling

sriramk117:
# Description

## Overview

This PR implements an ADA feeding perception node that enables semantic labeling. Semantic labeling acts as infrastructure in the perception code for advances to the user interface including: (1) allowing natural language prompting for ADA users, (2) auto-feeding features that give the robot the ability to continue to feed a specific food item on the plate until it is no longer there without user prompting, and (3) taking user preference (i.e. how they want to eat their meal) into account during bite selection.  

## ROS2 Interfaces

Firstly, the node creates an action server (GenerateCaption) that takes in a list of string labels describing the food items on the plate (for example, ['meat', 'cucumber']) and then, runs GPT-4o to compile these string labels into a visually-descriptive, sentence query for GroundingDINO (for example, "Food items including strips of grilled meat and seasoned cucumber spears arranged on a light gray plate."). 

The node also creates an action server (SegmentAllItems) that runs a pipeline consisting of a vision language model called GroundingDINO and the segmentation model EfficientSAM/SAM to output masks paired with semantic labels describing the masks for each of the detected food items on the plate. 

## Implementation Details

The perception node uses three foundation models to achieve semantic labeling: GPT-4o, GroundingDINO, and EfficientSAM/SegmentAnything.
 
From experimentation with prompting techniques on GroundingDINO and GroundingDINO's feature extraction functionality, we determined that prompting GroundingDINO with a sentence prompt that contains descriptive qualifiers results in significantly better performance for bounding box detection. Therefore, we query GPT-4o with the latest above plate image and a list of strings by the user labeling the food items on the plate to generate a visually-descriptive sentence incorporating those labels to describe the image. This sentence then gets used as input for GroundingDINO.

See the current system and user queries for GPT-4o below:

System Query: 
```
You are a prompt engineer that is assigned to describe items  
in an image you have been queried so that a vision language model 
can take in your prompt as a query and use it to for classification 
tasks. You respond in string format and do not provide any explanation 
for your responses.  
```

User Query: 
```
Your objective is to generate a sentence prompt that describes the food
items on a plate in an image. 
You are given an image of a plate with food items and a list of the food items 
on the plate.
Please compile the inputs from the list into a sentence prompt that effectively
lists the food items on the plate.
Add qualifiers to the prompt to better visually describe the food for the VLM
to identify. Don't add any irrelevant qualifiers.

Here is the input list of food items to compile into a string: {labels_list}

Here are some sample responses that convey how you should format your responses:

Food items including grapes, strawberries, blueberries, melon chunks, and 
carrots on a small, blue plate.

Food items including strips of grilled meat and seasoned cucumber 
spears arranged on a light gray plate.

Food items including baked chicken pieces, black olives, bell pepper slices, 
and artichoke on a plate.
```

GroundingDINO is a vision language model that detects objects in an image (through bounding boxes) given natural language prompts. We use GroundingDINO as it pairs semantic labels from the natural language prompt with each of the detected bounding boxes. 

Then, we use the segmentation model EfficientSAM/SAM to extract pixel-wise masks of each of the detected objects given their bounding boxes.

Here, we share a couple images of GroundingDINO's bounding box detection performance on various above plate image photos:

Screenshot 2024-11-09 at 2 07 59 AM

Screenshot 2024-11-09 at 2 05 26 AM

Screenshot 2024-11-09 at 2 06 43 AM

# Set Up Steps
Follow these steps to install necessary dependencies. : 
   - Open a new terminal window and go into your ADA workspace.
   - Go into the ada_feeding directory: `cd src/ada_feeding`
   - Run `pip install -r requirements.txt`

# Testing Procedure

### Test Setup
Before running any isolated tests, the following list of commands should be run each in separate terminal windows. Also, ensure that `source install/setup.bash` gets called in each of the windows that will run ROS commands.
- Launch the web app:
   - First, open three separate terminal tabs/windows and go into the correct folder in each window: `cd src/feeding_web_interface/feedingwebapp`
   - Run the following listed commands in the separate terminal tabs/windows:
      - Launch the web app server: `npm run start`
      - Set the web rtc server: `node --env-file=.env server.js`
      - Start the robot browser: `node start_robot_browser.js`
- Launch the perception nodes: `ros2 launch ada_feeding_perception ada_feeding_perception.launch.py`
- Launch the dummy nodes: `ros2 launch feeding_web_app_ros2_test feeding_web_app_dummy_nodes_launch.xml run_food_detection:=false run_face_detection:=false`
- Launch the nano bridge sender: `ros2 launch nano_bridge sender.launch.xml`
- Launch the nano bridge receiver: `ros2 launch nano_bridge receiver.launch.xml`

Below, see isolated tests for each of the two action servers that enable semantic labeling (the SegmentAllItems action and the GenerateCaption action). 

### Test GPT-4o Action:
- Open a separate terminal window. To send action goals to the GenerateCaption action server, use this command: `ros2 action send_goal /GenerateCaption ada_feeding_msgs/action/GenerateCaption "input_labels: ['placeholder']"`
   - In place of the placeholder, you would have a list of strings that concisely label the different food items to be eaten during the meal. Here is an example send goal ROS command you could use for this action: `ros2 action send_goal /GenerateCaption ada_feeding_msgs/action/GenerateCaption "input_labels: ['blueberries', 'carrots', 'strawberries', 'green grapes']"`
- If the action succeeds, you will find the result returned as a string called `caption`.

### Test GroundingDINO + SegmentAnything vision pipeline:
- Open a separate terminal window. To send action goals to the SegmentAllItems action server, use this command: `ros2 action send_goal /SegmentAllItems ada_feeding_msgs/action/SegmentAllItems "caption: Some input sentence query for the pipeline that visually describes the food items on the plate."`
   - Here is an example of a sample action goal you could send which includes an example input caption:  `ros2 action send_goal /SegmentAllItems ada_feeding_msgs/action/SegmentAllItems "caption: Food items on a blue plate including blueberries, carrots, strawberries, and green grapes."`
- If the action succeeds, it will return a list of Mask messages (containing information like the region of interest and a binary mask of the segmented item) named `detected_items` and a list of strings named `item_labels`
   - Each semantic label of the `item_labels` list corresponds to the Mask message at the same index of the `detected_items` list

I believe the most useful way to test the performance of semantic labeling is by merging these isolated tests together. First, send a goal to the GenerateCaption action server and get the resulting caption. Then send this caption as input to the SegmentAllItems action. This will be more identical to how the web interface sends goals to each of the action servers and fetches results to display during the bite selection stage.

### Testing Semantic Labeling on Different Images
An important part of testing is experimenting with various above plate images that have different food items and analyzing the performance of these perception nodes on these different images. To change the image that is used on the feeding web app dummy nodes (and hence, the one that gets passed as input into each of the action servers), follow these steps:
- Open the file `feeding_web_app_ros2_test/launch/feeding_web_app_dummy_nodes_launch.xml` in your favorite IDE:
- Pick an rgb image and its corresponding depth image from the folder `feeding_web_app_ros2_test/data`
- Change the image file names in the `rgb_path` and `depth_path` launch arguments accordingly
- Then follow the isolated tests detailed above to test semantic labeling on these images

# Before Merging
- [ ] `Squash & Merge`


.gitignore
```diff
@@ -2,6 +2,9 @@
 build/
 __pycache__/
 
+# Environment Variables file
+.env
+
```
jjaime2, line 7: I don't see a `.env` file added in this PR, but I'm guessing this was more for personal use. I'd recommend omitting this change unless it's relevant for the functionality of the PR.

ada_feeding/ada_feeding/behaviors/acquisition/compute_food_frame.py
```diff
@@ -336,7 +336,7 @@ def update(self) -> py_trees.common.Status:
             x_unit.vector, x_pos.vector
         )
 
-        # # If you need to send a fixed food frame to the robot arm, e.g., to 
+        # # If you need to send a fixed food frame to the robot arm, e.g., to
```
jjaime2: Unintended?

ada_feeding_msgs/action/SegmentAllItems.action
```diff
@@ -1,6 +1,8 @@
 # The interface for an action that gets an image from the camera and returns
 # the masks of all segmented items within that image.
 
+# The list of input semantic labels for the food items on the plate
+string caption
```
jjaime2, line 5: Comment seems misleading, I suspect this was an old comment for `item_labels`

requirements.txt
```diff
@@ -2,5 +2,8 @@ pyrealsense2
 overrides
 sounddevice
 scikit-spatial
+openai
+python-dotenv
```
jjaime2, line 6: Related to another comment on `.gitignore`. Are we using `.env` files for something?

ada_feeding_perception/config/republisher.yaml
```diff
@@ -3,27 +3,47 @@ republisher:
   ros__parameters:
     # The name of the topics to republish from
     from_topics:
+      - /camera/color/image_raw/compressed
+      - /camera/color/camera_info
+      - /camera/aligned_depth_to_color/image_raw/compressedDepth
+      - /camera/aligned_depth_to_color/camera_info
       - /local/camera/aligned_depth_to_color/image_raw/compressedDepth
```
jjaime2: What's the difference between the topic starting with `/local` and the ones you have added? Does republishing several streams of image data significantly increase the load on our network?

ada_feeding_perception/ada_feeding_perception/segment_all_items.py
```diff
@@ -0,0 +1,1279 @@
...
+
+# Standard imports
+import os
+import threading
+from typing import Optional, Tuple, Union
+
+# Third-party imports
+import cv2
+import time
+import random
```
jjaime2, line 16: Unused?

ada_feeding_perception/ada_feeding_perception/segment_all_items.py
```diff
@@ -0,0 +1,1279 @@
...
+        else:
+            self.initialize_sam(seg_model_name, seg_model_path)
+
+        # Convert between ROS and CV images
+        self.bridge = CvBridge()
+
+        # Create the shared resource to ensure that the action server rejects all
+        # goals while a goal is currently active.
+        self.active_goal_request_lock = threading.Lock()
+        self.active_goal_request = None
```
jjaime2, line 202: This used to be above what's currently on line 140. Since this involves a thread lock to reject other goals while this one is active, I'd double-check that this doesn't introduce unintended behavior.

ada_feeding_perception/ada_feeding_perception/segment_all_items.py
```diff
@@ -0,0 +1,1279 @@
...
+        API_KEY = os.getenv("OPENAI_API_KEY")
+        self.openai = OpenAI(api_key=API_KEY)
+
+    def read_params(
+        self,
+    ) -> Tuple[Parameter, Parameter, Parameter, Parameter, Parameter]:
+        """
+        Read the parameters for this node.
+
+        Returns
```
jjaime2, line 257: Probably should populate the return list

ada_feeding_perception/ada_feeding_perception/segment_all_items.py
```diff
@@ -0,0 +1,1279 @@
...
+        # Load the GroundingDINO model checkpoint
+        checkpoint = torch.load(groundingdino_model_path, map_location=self.device)
+        load_log = groundingdino.load_state_dict(
+            clean_state_dict(checkpoint["model"]), strict=False
+        )
+        self._node.get_logger().info(f"Loaded model checkpoint: {load_log}")
+        _ = groundingdino.eval()
+        self.groundingdino = groundingdino.to(device=self.device)
+
+        self._node.get_logger().info("...Done!")
```
jjaime2, line 497: Nitpick, can this be "Finished initializing GroundingDINO" or something? If I saw a "...Done!" message in my logs I would probably wonder where it came from if any other logs got published between.

ada_feeding_perception/ada_feeding_perception/segment_all_items.py
```diff
@@ -0,0 +1,1279 @@
...
+
+        Returns
+        -------
+        vlm_query: The caption generated by GPT-4o that is used as text input for
+                   GroundingDINO.
+        """
+        self._node.get_logger().info("Running GPT-4o...")
+
+        # Encode the image to JPEG format
+        _, buffer = cv2.imencode(".jpg", image)
```
jjaime2, line 657: Maybe check the `retval`. Not sure if this would ever fail but always nice to know when it does.

ada_feeding_perception/ada_feeding_perception/segment_all_items.py
```diff
@@ -0,0 +1,1279 @@
...
+        image_base64 = base64.b64encode(buffer).decode("utf-8")
+
+        # Write the system and user prompts for GPT-4o
+        system_query = f"""
+            You are a prompt engineer that is assigned to describe items  
+            in an image you have been queried so that a vision language model 
+            can take in your prompt as a query and use it to for classification 
+            tasks. You respond in string format and do not provide any explanation 
+            for your responses.  
+        """
```
jjaime2: I'm very curious how robust this is.

ada_feeding_perception/ada_feeding_perception/segment_all_items.py
```diff
@@ -0,0 +1,1279 @@
...
+        masks: The masks for each segmentation.
+        scores: The confidence scores for each segmentation.
+        """
+        self._node.get_logger().info("Segmenting image with EfficientSAM...")
+
+        # Convert image from BGR to RGB for Segment Anything
+        image = cv2.cvtColor(image, cv2.COLOR_BGR2RGB)
+
+        # Convert the image to a tensor
+        image_tensor = transforms.ToTensor()(image).to(self.device)
```
jjaime2, line 808: This can be `image_tensor = transforms.ToTensor()(image).to(device=self.device)` if we want consistency

ada_feeding_perception/ada_feeding_perception/segment_all_items.py
```diff
@@ -0,0 +1,1279 @@
...
+        # Convert input prompt (seed point or bounding box) to a tensor
+        if prompt == 0:
+            prompt_tensor = torch.tensor(np.array(seed_point).reshape((1, 1, 1, 2))).to(
+                device=self.device
+            )
+
+            # Define the labels for the input prompt
+            prompt_labels = torch.tensor([[[1]]]).to(device=self.device)
+        else:
+            prompt_tensor = torch.reshape(torch.tensor(bbox), [1, 1, 2, 2]).to(
```
jjaime2, line 819: Is the reshape to `[1, 1, 2, 2]` correct, or should this be `[1, 1, 1, 2]`?

ada_feeding_perception/ada_feeding_perception/segment_all_items.py
```diff
@@ -0,0 +1,1279 @@
...
+                center_x, center_y, w, h = box
+                # Get the bottom left and top right coordinates of the box
+                x0 = center_x - (w / 2)
+                y0 = center_y - (h / 2)
+                x1 = x0 + w
+                y1 = y0 + h
+                boxes_xyxy[phrase].append([x0, y0, x1, y1])
+
+        # Measure the elapsed time running GroundingDINO on the image prompt
+        inference_time = int(round((time.time() - inference_time) * 1000))
```
jjaime2, line 1080: Is this meant to be logged?

ada_feeding_perception/ada_feeding_perception/segment_all_items.py
```diff
@@ -0,0 +1,1279 @@
...
+        score: The confidence score outputed by the segmentation model for the mask.
+        mask: The pixel-wise mask detected.
+        image: The image the mask was detected on.
+        depth_img: The most recent depth image.
+        bbox: The bounding box from GroundingDINO.
+        """
+        # Calculate center of the bounding box and use it as the seed point for
+        # getting the connected component of the mask
+        center_x = (bbox[0] + bbox[2]) // 2
+        center_y = (bbox[1] + bbox[3]) // 2
```
jjaime2, line 1165: For this and other instances of `bbox`, does it have any properties like `xmin`, `xmax`, `ymin`, `ymax` to make this easily understandable?

ada_feeding_perception/ada_feeding_perception/segment_all_items.py
```diff
@@ -0,0 +1,1279 @@
...
+            width=int(bbox[2] - bbox[0]),
+            do_rectify=False,
+        )
+        mask_msg.mask = cv2_image_to_ros_msg(mask_img, compress=True)
+        mask_msg.rgb_image = cv2_image_to_ros_msg(cropped_image, compress=True)
+        mask_msg.depth_image = cv2_image_to_ros_msg(cropped_depth, compress=False)
+        mask_msg.average_depth = median_depth_mm / 1000.0
+        mask_msg.item_id = item_id
+        mask_msg.object_id = object_id
+        mask_msg.confidence = float(score)
```
jjaime2, line 1222: Old code for `SegmentFromPoint` didn't require a type-cast to `float`, it's fine if this is needed just wanted to check since I noticed the discrepancy.

ada_feeding_perception/ada_feeding_perception/segment_all_items.py
```diff
@@ -0,0 +1,1279 @@
...
+        Parameters
+        ----------
+        goal_handle: The goal handle for the action server.
+
+        Returns
+        -------
+        result: The result message containing masks for all food items detected in the image
+                paired with semantic labels.
+        """
+        self._node.get_logger().info("Received a new goal!")
```
jjaime2, line 1593: Should mention the `goal_handle` here

ada_feeding_perception/ada_feeding_perception/segment_all_items.py
```diff
@@ -0,0 +1,1279 @@
...
+            with self.active_goal_request_lock:
+                self.active_goal_request = None
+
+            # Cleanup the rate
+            cleanup()
+            return result
+
+        # Set the result after the task has been completed
+        self._node.get_logger().info("Goal not cancelled.")
+        self._node.get_logger().info("VIsion pipeline completed successfully.")
```
jjaime2: Also known as "Goal succeeded", also typo in "VIsion"

jjaime2 (commented):
Looking very good overall. Could likely use some cleanup but the core functionality seems to make sense. I would also check for `pylint` warnings if you haven't since I noticed many macros in the old code that this PR was based off of.

.gitignore
jjaime2: I noticed more references to a `env` file later in the code, where exactly does this come into play?

sriramk117: I'm adding environment variable functionality to our codebase so we can privately store API keys without exposing them publicly in github. In this particular case, it is for accessing the PRL OpenAI API key to invoke GPT-4o.

sriramk117: I'm assuming this may come in handy later on as well if we power perception w/ foundation models in the future.

ada_feeding_msgs/srv/GenerateCaption.srv
```diff
@@ -0,0 +1,11 @@
...
+# describing the food items on a plate and returns a sentence caption compiling
+# these labels used as a query for GroundingDINO detection.
+
+# A list of semantic labels corresponding to each of the masks of detected
+# items in the image
+string[] input_labels
+---
+# A sentence caption compiling the semantic labels used as a query for 
+# GroundingDINO to perform bounding box detections.
+string caption
```
amalnanavati: Nit: add newlines at the end of files. (I know not all files have it, but in general it is a best practice so we should enforce it on new/modified files)

ada_feeding_perception/config/republisher.yaml
amalnanavati: Hmm, this seems like the base branch of this PR has not been updated. Because all these extra topics got removed in a push a few months ago https://github.com/personalrobotics/ada_feeding/blob/ros2-devel/ada_feeding_perception/config/republisher.yaml . @sriramk117 if you pull the updated `main` and then rebase or merge on top of it, that should address this (non-)change

ada_feeding_perception/config/segment_all_items.yaml
```diff
@@ -0,0 +1,70 @@
...
+
+    # The threshold for bounding box detections by GroundingDINO
+    box_threshold: 0.30
+    # The threshold for text detections by GroundingDINO
+    text_threshold: 0.25
+
+    # A boolean to determine whether to visualize the bounding box predictions 
+    # made by GroundingDINO
+    viz_groundingdino: false
+
```
amalnanavati, line 70: Nit: newline.

I think your empty line at the end of the file has some whitespae, which is why Github doesn't recognize it as the empty line at the end of the file.

ada_feeding_perception/model/GroundingDINO_SwinT_OGC.py
amalnanavati, line 1: Nit: does this belong in `model` or `config`? This seems more like configuration. I believe what is downloaded to `model` is the actual files storing model weights.

ada_feeding_perception/ada_feeding_perception/segment_all_items.py
amalnanavati: Also, some of these, like `time` and `random`, are "standaard imports." `pylint` should help take care of some of these issues. Eventually, I want to set up `precommit` on this repo so reformatting and some level of linting happens automatically

ada_feeding_perception/ada_feeding_perception/segment_all_items.py
```diff
@@ -0,0 +1,1279 @@
...
+            self._node,
+            SegmentAllItems,
+            "SegmentAllItems",
+            self.execute_callback,
+            goal_callback=self.goal_callback,
+            cancel_callback=self.cancel_callback,
+            callback_group=MutuallyExclusiveCallbackGroup(),
+        )
+
+        # If the GroundingDINO results visualization flage is set, then a publisher
```
amalnanavati, line 240: Nit: typo `flag`

ada_feeding_perception/ada_feeding_perception/segment_all_items.py
amalnanavati: On quick glance, I don't think it should cause any issues because it is only used in the action server callbacks, so should be fine as long as it is created before the action server (which it is). But +1 to Jose's feedback to double-check, just to be safe.

ada_feeding_perception/ada_feeding_perception/segment_all_items.py
amalnanavati: And if you're changing that, I'd recommend changing it in the intiialization of (efficient)SAM as well, for consistency. (I agree that Jose's suggestion is better since its more descriptive)

ada_feeding_perception/ada_feeding_perception/segment_all_items.py
```diff
@@ -0,0 +1,1279 @@
...
+        sam.to(device=self.device)
+
+        # Create the predictor
+        # NOTE: If we allow for concurrent goals, this should be protected by
+        # a lock.
+        self.sam = SamPredictor(sam)
+
+        self._node.get_logger().info("...Done!")
+
+    def initialize_efficient_sam(self, model_name: str, model_path: str) -> None:
```
amalnanavati, line 536: Since this is the second file that uses these exact functions (for (Efficient)SAM), can you instead add these functions to a/the helpers file? To avoid copy-pasted code.

ada_feeding_perception/ada_feeding_perception/segment_all_items.py
```diff
@@ -0,0 +1,1279 @@
...
+            Food items including strips of grilled meat and seasoned cucumber 
+            spears arranged on a light gray plate.
+
+            Food items including baked chicken pieces, black olives, bell pepper slices, 
+            and artichoke on a plate.
+        """
+
+        # Run GPT-4o to generate a sentence caption given the user query, system query,
+        # and image prompt
+        response = self.openai.chat.completions.create(
```
amalnanavati, line 696: Services ideally should not block. However, in this case, I imagine if the API response is not received, it is possible that the code will hang here, right?

I'd recommend adding a `timeout` to the `run_gpt4o` function (and maybe even to the service msg), and to do an asynchronous query to ChatGPT instead. That way, even if something is wrong with ChatGPT, the service call will not get stuck.

amalnanavati: Also, you should do appropriate error checking on the response. I imagine there are various error cases that could arise (check the API for details), and we need to properly handle those.

ada_feeding_perception/ada_feeding_perception/segment_all_items.py
```diff
@@ -0,0 +1,1279 @@
...
+                                "url": f"data:image/jpeg;base64,{image_base64}"
+                            },
+                        },
+                    ],
+                },
+            ],
+        )
+
+        # Get the caption generated by GPT-4o
+        vlm_query = response.choices[0].message.content
```
amalnanavati, line 716: Related to error handling, is it possible that `response` is None, or that `response.choices` is length 0? You need to handle cases like that.

ada_feeding_perception/ada_feeding_perception/segment_all_items.py
```diff
@@ -0,0 +1,1279 @@
...
+
+        return vlm_query
+
+    def run_sam(
```
amalnanavati, line 735: If this is the same as / similar to `SegmentFromPoint`, this should be in a `helper` function.

If the default `helpers` file is getting too long, you can make a new file like `segmentation_helpers.py`

amalnanavati: Same with `run_efficient_sam` below.

ada_feeding_perception/ada_feeding_perception/segment_all_items.py
amalnanavati: Adding on to Jose's point, adding a comment (here and above) explaining the reshape will make it more readable

ada_feeding_perception/ada_feeding_perception/segment_all_items.py
```diff
@@ -0,0 +1,1279 @@
...
+        image: The CV2 above the plate image to retrieve semantically labeled bounding
+               boxes from.
+        caption: The caption to use as text input for GroundingDINO.
+        box_threshold: The threshold for the bounding box.
+        text_threshold: The threshold for the text.
+
+        Returns
+        -------
+        bbox_predictions: A dictionary containing the bounding boxes for each food item label
+                        detected from the image.
```
amalnanavati, line 1004: Add here how bounding boxes are represented. I'm guessing it is `(x, y, w, h)` where `x,y` is the top-left corner?

ada_feeding_perception/ada_feeding_perception/segment_all_items.py
amalnanavati: Also, why are you multiplying by 1000?

In general, I find it more readable to specify the units of measurements after the variable name itself, e.g., in this case maybe it should be `inference_time_ms` given the factor of `1000`?

ada_feeding_perception/ada_feeding_perception/segment_all_items.py
```diff
@@ -0,0 +1,1279 @@
...
+        """
+        self._node.get_logger().info("Running GroundingDINO...")
+
+        # Set the initial time to measure the elapsed time running GroundingDINO on the
+        # desired image and text prompts.
+        inference_time = time.time()
+
+        # Convert image to Image pillow
+        image_pil, image_transformed = self.load_image(image)
```
amalnanavati: image_pil seems unused -- any reason you're returning/storing it?

amalnanavati: Also, `load_image` is quite unclear as a name imo. I'd suggest something like `convert_cv2_to_tensor`.

ada_feeding_perception/ada_feeding_perception/segment_all_items.py
```diff
@@ -0,0 +1,1279 @@
...
+        image: The image in tensor format.
+        """
+        # Convert image from BGR to RGB to convert CV2 image to image pillow
+        image_array = cv2.cvtColor(image_array, cv2.COLOR_BGR2RGB)
+
+        # Convert image to image pillow to apply transformation
+        image_pil = ImagePIL.fromarray(image_array, mode="RGB")
+        transform = T.Compose(
+            [
+                T.RandomResize([800], max_size=1333),
```
amalnanavati, line 1130: What all transforms do you do? To make this function more general, I'd suggest you make them parameters, with default values. (i.e., it looks like you're resizing the image to size 800x800, but maybe in the future we may want to resize the image to a different size).

ada_feeding_perception/ada_feeding_perception/segment_all_items.py
```diff
@@ -0,0 +1,1279 @@
...
+                T.ToTensor(),
+                T.Normalize([0.485, 0.456, 0.406], [0.229, 0.224, 0.225]),
```
amalnanavati, line 1132: What are these magic numbers? They should at least have a comment explaining them, if not be moved to parameters of this function outright

ada_feeding_perception/ada_feeding_perception/segment_all_items.py
amalnanavati: You coudl consider using a [namedtuple](https://docs.python.org/3/library/collections.html#collections.namedtuple) for this

ada_feeding_perception/ada_feeding_perception/segment_all_items.py
amalnanavati: Also, these small changes are precisely why its best to not copy-paste code (e.g., if casting to `float` improves the code here, we should ideally percolate it back to `SegmentFromPoint`). So this is one more of the functions that should be in a helper.

Another thing you can consider is making `SegmentAllItems` **inherit** from the `SegmentFromPoint` class, and only override some functions.

ada_feeding_perception/ada_feeding_perception/segment_all_items.py
```diff
@@ -0,0 +1,1279 @@
...
+        bbox_predictions = self.run_grounding_dino(
+            image, caption, self.box_threshold, self.text_threshold
+        )
+
+        # Publish a visualization of the GroundingDINO predictions, if the visualization
+        # flag is set to true
+        if self.viz_groundingdino:
+            self.visualize_groundingdino_results(image, bbox_predictions)
+
+        # Collect the top contender mask for each food item label detected by
```
amalnanavati, line 1352: I'd suggest having a parameter to the function (set by default to `1`) that controls how many of the top masks you use per food item

ada_feeding_perception/ada_feeding_perception/segment_all_items.py
```diff
@@ -0,0 +1,1279 @@
...
+                        item_id, phrase, scores[0], masks[0], image, depth_img, box
+                    )
+                    detected_items.append(mask_msg)
+                    item_labels.append(phrase)
+
+        result.detected_items = detected_items
+        result.item_labels = item_labels
+
+        # Measure the elapsed time running GroundingDINO on the image prompt
+        inference_time = int(round((time.time() - inference_time) * 1000))
```
amalnanavati, line 1386: Same comment as above re. ms vs sec and adding units to variable names

ada_feeding_perception/ada_feeding_perception/segment_all_items.py
```diff
@@ -0,0 +1,1279 @@
...
+        # Check if the image and camera info are available
+        if latest_img_msg is None or camera_info is None:
+            self._node.get_logger().error("Image or camera info not available.")
+            return SegmentAllItems.Result()
+
+        # Get the caption from the goal request
+        caption = goal_handle.request.caption
+
+        # Create a rate object to control the rate of the vision pipeline
+        rate = self._node.create_rate(self.rate_hz)
```
amalnanavati, line 1614: I recently learned that rates live forever in a node unless explicitly deleted by `destroy_rate` (see comment [here](https://github.com/ros2/rclpy/blob/78f5e14cdabeb047ab39829caef7a4872f3179ff/rclpy/rclpy/node.py#L1891)). So I'd recommend adding that at the end of this function, else the rate will keep taking up callback resources even after the action has finished.

ada_feeding_perception/ada_feeding_perception/segment_all_items.py
```diff
@@ -0,0 +1,1279 @@
...
+            self.run_vision_pipeline, latest_img_msg, caption
+        )
+
+        # Wait for the vision pipeline to finish and keep publishing
+        # feedback (elapsed time) while waiting
+        feedback = SegmentAllItems.Feedback()
+        while (
+            rclpy.ok()
+            and not goal_handle.is_cancel_requested
+            and not vision_pipeline_task.done()
```
amalnanavati, line 1631: Are there cases where the vision pipeline could hang? If so, I'd recommend adding a timeout to the action (maybe in the action message itself) to be robust to that

ada_feeding_perception/ada_feeding_perception/segment_all_items.py
```diff
@@ -0,0 +1,1279 @@
...
+
+        # If there is a cancel request, cancel the vision pipeline task
+        if goal_handle.is_cancel_requested:
+            self._node.get_logger().info("Goal cancelled.")
+            goal_handle.canceled()
+            result = SegmentAllItems.Result()
+            result.status = result.STATUS_CANCELLED
+
+            # Clear the active goal
+            with self.active_goal_request_lock:
```
amalnanavati, line 1647: Why not do this in the cleanup function itself? (both here and below)

PR #209, opened 2024-12-18
fix typo in Setup (System Configuration) in README

yewon-lee:
# Description

In the original README, it reads "Our system includes two computers, lovelace (Lenovo Legion 5i) and nano (Nvidia Jetson Nano), where the former runs camera code and the latter runs everything else". 

However, this is incorrect, as nano runs the camera code and lovelace runs everything else. I have updated the README to correct this.

# Before Merging

- \[ \] `Squash & Merge`

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ This README is the definitive source for downloading, installing, and running th
 
 ### Setup (System Configuration)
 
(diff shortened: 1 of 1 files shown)


amalnanavati (approved):
LGTM. Thanks!

PR #210, opened 2024-12-20
Move AcquisitionReport work into asynch thread

amalnanavati:
# Description

This PR takes a change that was made for deployment (see this commit) and merges it into the overall repo. Specifically, this change makes AcquisitionReport asynchronous; it will always return success to the web app, and will actually perform the online learning update in an asynchronous thread. This means that the web app no longer has to wait for the policy update to succeed, which could take longer depending on the number of trainable parameters.

# Testing procedure

- [x] Tested via the home deployment.
- [x] In addition, before merging in, we should verify that it works on `lovelace` (e.g., the web app succesfully invokes an acquisition report, without blocking, and the data gets saved)

- \[x\] `pre-commit run --all-files`
- \[x\] Run your code through [pylint](https://pylint.readthedocs.io/en/latest/). `pylint --recursive=y --rcfile=.pylintrc .`. All warnings but `fixme` must be addressed.

# Before Merging

- \[ \] `Squash & Merge`

### DIFF
--- a/ada_feeding_action_select/ada_feeding_action_select/policy_service.py
+++ b/ada_feeding_action_select/ada_feeding_action_select/policy_service.py
@@ -381,7 +381,6 @@ class PolicyServices(Node):
         # Start the asynch thread
         request_copy = copy.deepcopy(request)
         response_copy = copy.deepcopy(response)
-        # self.report_callback_work(request_copy, response_copy)
         thread = threading.Thread(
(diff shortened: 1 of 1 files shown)


amalnanavati:
Tested on `lovelace` and verified it worked.

This code was also used as part of the 2024 deployment, so I'm extra-confident that it works :)

PR #216, opened 2025-02-19
Map weights to CUDA or CPU depending on which is enabled

jjaime2:
# Description

This PR adds a check for whether CUDA is enabled before loading our checkpoint weight files. If it isn't enabled, we map the weights to CPU instead. This is particularly relevant since the current instructions for setting up machines to run `ada_feeding` doesn't explicitly require enabling CUDA.

# Testing procedure

1. Launch the feeding software on machines both with and without CUDA enabled, e.g. `lovelace` and one of the lab machines
2. Observe that there are no errors related to loading weights due to incompatible CUDA configurations

- \[x\] `pre-commit run --all-files`
- \[x\] Run your code through [pylint](https://pylint.readthedocs.io/en/latest/). `pylint --recursive=y --rcfile=.pylintrc .`. All warnings but `fixme` must be addressed.

# Before Merging

- \[ \] `Squash & Merge`

### DIFF
--- a/ada_feeding_action_select/ada_feeding_action_select/adapters/hapticnet_adapter.py
+++ b/ada_feeding_action_select/ada_feeding_action_select/adapters/hapticnet_adapter.py
@@ -47,10 +47,13 @@ class HapticNetPosthoc(PosthocAdapter):
 
         # Init CUDA
         self.use_cuda = torch.cuda.is_available()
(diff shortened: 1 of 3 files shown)


sriramk117 (approved):
PR makes sense and code looks good to me

